PR #1, opened 2018-12-22
Add block functionality.  Provide README example.

finish06:
Unblock functionality will require discussion as the device is no longer registered as a client once unblocked.  Will possibly need to be a class function, not an object function?

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ Asynchronous library to communicate with Unifi Controller
 ## Acknowledgements
 * Paulus Schoutsen (balloob) creator of aiohue which most of this code repository is modeled after.
 
-#### Exampl
+#### Example
 ```python
 import asyncio
 import aiohttp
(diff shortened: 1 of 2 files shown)


finish06:
Maybe wrap the unblock function with `@staticmethod` so it is within the `Client` class, however does not require an object instantiation.

Kane610:
Sorry that its taken some time to respond. I'm slowly getting back into it after Christmas vacation.

I have a couple of bugs to fix for UniFi so I start with those to get back to it and then look at this.

Kane610:
I've started to think more about this. This will need to be a part of the controller. Since there is no client created for this and thus no reference to request, it is better to keep block and unblock on a controller level.

Some thought into how this will be handled on hass side is also necessary, since there is no support for configuration.yaml with UniFi component some other mean of configuration is required. I have a WIP on adding options to config entries, I'd say a page to register entities that can blocked and a page to unregister entities that can be blocked.

What are you're thoughts?

finish06:
I think handling it via the controller is a great idea.  Much better then via the client approach I initially considered.

Kane610:
Solved by #4 

Sorry it took so long to get this done

PR #4, opened 2019-07-25
Block client support

Kane610:


Kane610:
Like https://github.com/Kane610/aiounifi/pull/1 but different

PR #14, opened 2020-02-25
Support new url_base used by UDM

Kane610:


aiounifi/controller.py
```diff
@@ -58,18 +61,34 @@ def __init__(
         self.devices = None
         self.wlans = None
 
+    async def check_unifi_os(self):
+        self.is_unifi_os = True
```
JohnLahr: `is_unifi_os` is already initialized as `False`, so won't this line incorrectly change it to `True` regardless of the try block below? You're also not setting this value back to `False` if it fails, so this value will always remain `True` as soon as `check_unifi_os` is called.

It also makes the test on line 72 redundant, which makes me wonder if this line was supposed to be `False` instead of `True`.

Kane610: It has no real effect, I did that to try out everything with the old controller

PR #17, opened 2020-04-01
Unifi Dream Machine Pro 404s on Login

JohnLahr:
- Removed duplicate loop (implicit when using run() in Python3.7+)
- Added missing x-csrf-token header
- Added null-safe property access to raise_on_error method

### DIFF
--- a/aiounifi/__main__.py
+++ b/aiounifi/__main__.py
@@ -89,8 +89,8 @@ async def main(host, username, password, port, site, sslcontext=False):
 
 
 if __name__ == "__main__":
-    # logging.basicConfig(format="%(message)s", level=logging.DEBUG)
-    logging.basicConfig(format="%(message)s", level=logging.INFO)
+    logging.basicConfig(format="%(message)s", level=logging.DEBUG)
+    # logging.basicConfig(format="%(message)s", level=logging.INFO)
     parser = argparse.ArgumentParser()
     parser.add_argument("host", type=str)
     parser.add_argument("username", type=str)
--- a/aiounifi/__main__.py
+++ b/aiounifi/__main__.py
@@ -31,6 +31,7 @@ async def unifi_controller(
 
     try:
         with async_timeout.timeout(10):
+            # await controller.is_unifi_os()
             await controller.login()
         return controller
 
@@ -41,7 +42,7 @@ async def unifi_controller(
         LOGGER.warning(f"Connected to UniFi at {host} but not registered")
 
     except (asyncio.TimeoutError, aiounifi.RequestError):
-        LOGGER.error(f"Error connecting to the UniFi controller at {host}")
+        LOGGER.exception(f"Error connecting to the UniFi controller at {host}")
 
     except aiounifi.AiounifiException:
         LOGGER.exception("Unknown UniFi communication error occurred")
--- a/aiounifi/__main__.py
+++ b/aiounifi/__main__.py
@@ -34,6 +34,9 @@ async def unifi_controller(
             await controller.login()
         return controller
 
+    except aiounifi.LoginRequired:
+        LOGGER.warning(f"Connected to UniFi at {host} but couldn't log in")
+
     except aiounifi.Unauthorized:
         LOGGER.warning(f"Connected to UniFi at {host} but not registered")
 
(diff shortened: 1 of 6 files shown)


aiounifi/__main__.py
```diff
@@ -51,10 +51,9 @@ def signalling_callback(signal, data):
 async def main(host, username, password, port, site, sslcontext=False):
     """Main function."""
     LOGGER.info("Starting aioUniFi")
-    loop = asyncio.get_event_loop()
```
JohnLahr, line 54: This is done inside the `asyncio.run()` method already, so it is redundant here. (In fact, it raises exceptions here for me on Python 3.7.7, so I'm not even sure how you were running this before...)

aiounifi/controller.py
```diff
@@ -61,8 +61,18 @@ def __init__(
         self.devices = None
         self.wlans = None
 
+        self.headers = None
```
JohnLahr, line 64: I dunno how to use the cookie jar, so this is a sloppy way of getting around that for now.

aiounifi/controller.py
```diff
@@ -61,8 +61,18 @@ def __init__(
...
+
     async def check_unifi_os(self):
         response = await self.request("get", include_site=False)
+
+        headers = {
```
JohnLahr, line 69: These headers are set in my browser by response headers in the original login request if I don't already have a cookie, or if my cookie has expired. Both of these headers have to be set in order for the UDM Pro to accept the request. If either of these headers are missing, I would get the same 404 you were experiencing.

I don't know much about the `aiohttp.CookieJar`, but I feel like these values should have been included in that.

aiounifi/controller.py
```diff
@@ -179,7 +188,7 @@ def message_handler(self, message: dict) -> dict:
                 if res.content_type == "application/json":
                     response = await res.json()
                     _raise_on_error(response)
-                    return response["data"]
+                    return response.get("data") or response
```
JohnLahr, line 191: I noticed not every request included a `data` object in the response, but expected the entire response instead.

aiounifi/controller.py
```diff
@@ -190,5 +199,10 @@ def message_handler(self, message: dict) -> dict:
 
 def _raise_on_error(data):
     """Check response for error message."""
-    if isinstance(data, dict) and data["meta"]["rc"] == "error":
-        raise_error(data["meta"]["msg"])
+    if isinstance(data, dict):
```
JohnLahr, line 202: Again, maybe there's a difference in our setups, but mine raised exception without null-safe property access. Also noticed not every error response came in the format `"data.meta.rc": "error"`, so this catches the other format I experienced (_i.e._ invalid password)

aiounifi/__main__.py
Kane610: I think Im running 3.7.6, no exceptions here :)

aiounifi/controller.py
Kane610: This didn't work for me since it didn't cover the case when the data value is an empty list, but using parts of it. thanks

aiounifi/controller.py
Kane610: Those other errors are probably UniFi os specific errors. How did you get them?

aiounifi/controller.py
Kane610: Thanks, this did the trick

aiounifi/controller.py
JohnLahr: Most of them were on login. Using bad credentials, trying to log into an account with 2FA enabled--things like that. Didn't keep track of which ones specifically, but I do remember logging in with bad credentials returned a response with an `Errors` array and no `meta` object.

Kane610: So many cases to cover. Long term I should probably rewrite parts of it...

JohnLahr: So unfortunate that they don't provide documentation on their API...

Kane610: You don't happen to have a raw example of this?

JohnLahr: I don't.. Going back through it though, it looks like a commit you had pushed just before I opened this PR had already fixed that--at least with invalid logins. If I rebase off the commit after this one, I don't have that issue any longer.

Kane610: Well great then. Thanks for your help! I've reflected your changes in the already merged PR, so I'll close this one.

Kane610:
All suggested changes handled in #14

Issue #24, opened 2020-04-28
EVT_WU_RoamRadio should be included in list of wireless client events


cameronr:
It doesn't look common, but I received these events together with no intervening `EVT_WU_Connected` event:

> 2020-04-27 19:58:51 DEBUG (MainThread) [aiounifi.websocket] {"meta":{"rc":"ok","message":"events"},"data":[{"user":"<<user_mac>>","ssid":"<<mac>>","hostname":"<<hostname>>","ap":"<<ap_mac>>","duration":477,"bytes":19130891,"key":"EVT_WU_Disconnected","subsystem":"wlan","site_id":"<<site_id>>","time":1588042724000,"datetime":"2020-04-28T02:58:44Z","msg":"User[<<user_mac>>] disconnected from \"<<mac>>\" (7m 57s<<site_id>> connected, 18.24M bytes, last AP[<<ap_mac>>])","_id":"5ea79be852faff00f29407d2"},{"ap":"<<ap_mac>>","user":"<<user_mac>>","radio_from":"na","radio_to":"ng","channel_from":"36","channel_to":"11","key":"EVT_WU_RoamRadio","subsystem":"wlan","site_id":"<<site_id>>","time":1588042728065,"datetime":"2020-04-28T02:58:48Z","msg":"User[<<user_mac>>] roams from \"channel 36(na)\" to \"channel 11(ng)\" at AP[<<ap_mac>>]","_id":"5ea79be852faff00f29407d4"}]}

The result in Home Assistant is that `wireless_connection` was False, preventing the clearing of `cancel_scheduled_update`, causing my device to be set as `not_home`.

I think `EVT_WU_RoamRadio` should be included as a wireless client event so it can be picked up in HA.

PR #25, opened 2020-04-28
Add radio roaming events:

cameronr:
EVT_WU_RoamRadio (wireless user changing radios)
EVT_WG_RoamRadio (wireless guest changing radios)

Addresses #24 so downstream code in HA can be fixed.


setup.py
```diff
@@ -13,7 +13,7 @@
 setup(
     name="aiounifi",
     packages=["aiounifi"],
-    version="18",
+    version="19",
```
Kane610, line 16: Bumping version doesn't belong to a feature PR

```suggestion
    version="18",
```

cameronr: ah, my bad. fixed

Issue #43, opened 2020-10-27
Add ability to Enable/Disable DPI restrictions


hunterjm:
We recently caught my daughter watching YouTube on her Chromebook while supposed to be in school, I created a DPI filter to block streaming media on the VLAN her machine is connected to. It works well, but would be nice to enable Home Assistant to import these configurations and expose a switch to enable/disable the block so I can create automations around it. Happy to open a PR eventually (unless you want to take a stab at it), but figured I would create this issue with the API details to start.

### The below are called together to fetch the DPI Restriction Definitions:

**`GET /api/s/default/rest/dpigroup`**
```json
{
	"meta": {
		"rc": "ok"
	},
	"data": [{
		"_id": "5ba29dd8e3c58f026e9d7c4a",
		"attr_no_delete": true,
		"attr_hidden_id": "Default",
		"name": "Default",
		"site_id": "5ba29dd4e3c58f026e9d7c38"
	}, {
		"_id": "5f976f4ae3c58f018ec7dff6",
		"name": "No Media",
		"site_id": "5ba29dd4e3c58f026e9d7c38",
		"dpiapp_ids": ["5f976f62e3c58f018ec7e17d"]
	}]
}
```

**`GET /api/s/default/rest/dpiapp`**
```json
{
	"meta": {
		"rc": "ok"
	},
	"data": [{
		"_id": "5f976f62e3c58f018ec7e17d",
		"apps": [],
		"blocked": true,
		"cats": ["4"],
		"enabled": true,
		"log": true,
		"site_id": "5ba29dd4e3c58f026e9d7c38"
	}]
}
```

### The below are called together to update the DPI Restriction Definitions:

**`PUT /api/s/default/rest/dpigroup/{dpigroup_id}`**
**Request**
```json
{
	"_id": "5f976f4ae3c58f018ec7dff6",
	"name": "No Media",
	"site_id": "5ba29dd4e3c58f026e9d7c38",
	"dpiapp_ids": ["5f976f62e3c58f018ec7e17d"]
}
```
**Response**
```json
{
	"meta": {
		"rc": "ok"
	},
	"data": []
}
```

**`PUT /api/s/default/rest/dpiapp/{dpiapp_id}`**
**Request**
```json
{
	"_id": "5f976f62e3c58f018ec7e17d",
	"apps": [],
	"blocked": true,
	"cats": ["4"],
	"enabled": false,
	"log": true,
	"site_id": "5ba29dd4e3c58f026e9d7c38"
[...]

PR #44, opened 2020-10-27
Add ability to enable/disable DPI Restrictions

hunterjm:
Implements #43


hunterjm:
Changes likely incoming as I work on the Home Assistant implementation.

aiounifi/dpi.py
```diff
@@ -0,0 +1,76 @@
...
+            app_url = f"{APP_URL}/{app_id}"
+            data = {"enabled": True}
+            await self._request("put", app_url, json=data)
+
+    async def async_disable(self, dpi: DPIRestrictionGroup) -> None:
+        """Disable DPI Restriction Group Apps."""
+        for app_id in dpi.dpiapp_ids:
+            app_url = f"{APP_URL}/{app_id}"
+            data = {"enabled": False}
+            await self._request("put", app_url, json=data)
```
Kane610, line 76: is it the expected way to manage all apps per group?

If yes, can we do a gather on the calls?
```suggestion
    async def async_enable(self, dpi: DPIRestrictionGroup) -> None:
        """Enable DPI Restriction Group Apps."""
        calls = []
        for app_id in dpi.dpiapp_ids:
            app_url = f"{APP_URL}/{app_id}"
            data = {"enabled": True}
            calls.append(self._request("put", app_url, json=data))
        await asyncio.gather(*calls)

    async def async_disable(self, dpi: DPIRestrictionGroup) -> None:
        """Disable DPI Restriction Group Apps."""
        calls = []
        for app_id in dpi.dpiapp_ids:
            app_url = f"{APP_URL}/{app_id}"
            data = {"enabled": False}
            calls.append(self._request("put", app_url, json=data))
        await asyncio.gather(*calls)
```

aiounifi/dpi.py
```diff
@@ -0,0 +1,76 @@
...
+
+    @property
+    def enabled(self) -> bool:
+        return self.raw["enabled"]
+
+
+class DPIRestrictionGroups(APIItems):
+    """Represents DPI Group configurations."""
+
+    KEY = "name"
```
Kane610, line 45: Can a user rename the app at any time? Then maybe the _id should be used?

Kane610 (commented):
Overall looks good, couple of comments to think about

aiounifi/dpi.py
hunterjm: Fair enough. This was taken from WLANs.

aiounifi/dpi.py
hunterjm: The frontend implementation is a little strange in the Controller currently. You can't enable/disable a group, but there's no reason to have more than one app per group because the categories are an array. I guess you could enable/disable a specific set of categories independently, but there is no "name" attribute on the apps, so I went with a simplistic implementation which mandates control of all apps in a group being the same.

Kane610: Fair enough 👍🏻

aiounifi/dpi.py
Kane610: Good point, though WLANs is not in use :)

Kane610 (approved):
Awesome!

PR #51, opened 2021-01-05
Support client reconnect command

rfleming71:
Support forcing a client to reconnect to the network.
The command is only valid for wireless clients.

### DIFF
--- a/aiounifi/clients.py
+++ b/aiounifi/clients.py
@@ -26,6 +26,11 @@ class Clients(APIItems):
         data = {"mac": mac, "cmd": "unblock-sta"}
         await self._request("post", URL_CLIENT_STATE_MANAGER, json=data)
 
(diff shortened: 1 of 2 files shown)


aiounifi/clients.py
```diff
@@ -26,6 +26,11 @@ async def async_unblock(self, mac: str) -> None:
         data = {"mac": mac, "cmd": "unblock-sta"}
         await self._request("post", URL_CLIENT_STATE_MANAGER, json=data)
 
+    async def async_reconnect(self, mac: str) -> None:
+        """Force a client to reconnect to the network."""
```
Kane610, line 30:
```suggestion
        """Force a wireless client to reconnect to the network."""
```

Kane610 (approved):
Thanks for your contribution!

PR #61, opened 2021-05-17
Use mypy in github actions

Kane610:


aiounifi/dpi.py
```diff
@@ -71,7 +47,7 @@ def site_id(self) -> str:
         return self.raw["site_id"]
 
 
-class DPIRestrictionApps(DPIItems):
+class DPIRestrictionApps(APIItems):
```
hunterjm, line 50: I believe this is the culprit for https://github.com/home-assistant/core/issues/57845

`DPIItems` processes state differently than `APIItems`.

aiounifi/dpi.py
```diff
@@ -134,14 +108,14 @@ def dpiapp_ids(self) -> List[str]:
...
+    def enabled(self) -> Union[DPIRestrictionApps, None, bool]:
         """Are all apps in group enabled."""
         return self.apps and all(
-            [self.apps[id].enabled for id in self.apps if id in self.dpiapp_ids]
+            [self.apps[id].enabled for id in self.apps if id in self.dpiapp_ids]  # type: ignore
         )
 
 
-class DPIRestrictionGroups(DPIItems):
+class DPIRestrictionGroups(APIItems):
```
hunterjm, line 118: Here too

aiounifi/dpi.py
Kane610: Thanks! I will have a look tonight!

aiounifi/dpi.py
Kane610: Double thanks @hunterjm !

PR #70, opened 2021-09-07
Fix NPE when x-csrf-token is not present in API response

edwardmp:
I ran into an issue where the Unifi controller API doesn't seem to return a CSRF header (not sure why, but apparently we need to deal with it).

If that's the case, we will still set the header with a None value in `controller.py`. This in turn leads to an error being thrown in `aiohttp._http_writer._serialize_headers` later on, because it essentially causes an NPE trying to serialize an empty header value (which of course, makes no sense).

This change ensures the header is not set if it's not in the response, thereby solving the NPE. 

Stack trace before the fix was applied:
```
  File "/usr/local/lib/python3.9/site-packages/aiounifi/controller.py", line 108, in login
    await self._request("post", url=url, json=auth)
  File "/usr/local/lib/python3.9/site-packages/aiounifi/controller.py", line 274, in _request
    async with self.session.request(
  File "/usr/local/lib/python3.9/site-packages/aiohttp/client.py", line 1117, in __aenter__
    self._resp = await self._coro
  File "/usr/local/lib/python3.9/site-packages/aiohttp/client.py", line 542, in _request
    resp = await req.send(conn)
  File "/usr/local/lib/python3.9/site-packages/aiohttp/client_reqrep.py", line 668, in send
    await writer.write_headers(status_line, self.headers)
  File "/usr/local/lib/python3.9/site-packages/aiohttp/http_writer.py", line 118, in write_headers
    buf = _serialize_headers(status_line, headers)
  File "aiohttp/_http_writer.pyx", line 137, in aiohttp._http_writer._serialize_headers
  File "aiohttp/_http_writer.pyx", line 109, in aiohttp._http_writer.to_str
TypeError: Cannot serialize non-str key None
```

### DIFF
--- a/aiounifi/controller.py
+++ b/aiounifi/controller.py
@@ -92,7 +92,9 @@ class Controller:
         response = await self._request("get", url=self.url, allow_redirects=False)
         if response.status == 200:
             self.is_unifi_os = True
-            self.headers = {"x-csrf-token": response.headers.get("x-csrf-token")}
+            # the header isn't always set, prevent it being set to none, as it would cause a NPE when serializing the headers elsewhere 
+            if response.headers.get("x-csrf-token") is not None:
+                self.headers = {"x-csrf-token": response.headers.get("x-csrf-token")}
 
     async def login(self) -> None:
         """Log in to controller."""


edwardmp:
By the way, noticed [this](https://github.com/jens-maus/node-unifi/blob/41fe6d7637afdbec6d5c932dfe4a0be1566c61b2/unifi.js#L2963) Node library for Unifi Controller also has similar handling for the CSRF token when it's not set, implying this is indeed a scenario we need to be able to gracefully handle.

Kane610:
Thanks! I will have a look on Sunday or early next week

Kane610:
A difference to the nodejs implementation is that if the header is not available it expects the controller to be non-unifi. This change doesnt change the fact that it is still expected to be unifi os.

Will communication with unifios really work without the token?

edwardmp:
I'm not an expert w.r.t. Unifi, I just noticed this didn't work out of the box when I used the Unifi integration for [home assistant](https://www.home-assistant.io/integrations/unifi/) which uses this lib under the hood.

In my case I have a Unifi Dream Machine Pro, and I debugged the auth request and no CSRF header is returned in the response. I'll do some more digging when I have time, all I can say is this works for me right now.

Kane610:
Ok, I guess the change doesn't hurt.

Kane610 (changes requested):
Fix the linting and add a test to https://github.com/Kane610/aiounifi/blob/master/tests/test_controller.py

edwardmp:
Hey @Kane610 

I think I know why this issue is happening.. Tried another library that actually had the same issue, see [this GH issue](https://github.com/briis/unifiprotect/issues/312).

TLDR: when running aiounifi on the dream machine itself, the header simply doesn't get served.

Would you be in favor of adding a config option to force assume the Unifi device uses UnifiOS to circumvent this?
Otherwise, the only proper way to detect whether a decice runs UnifiOS would probably be to try the UnifiOS auth URL first, then if doesn't exist it means the device doesn't run UnifiOS.

Any thoughts?

Kane610:
I would like to keep any knowledge about the connection in aiounifi

Kane610:
What do you mean with "TLDR: when running aiounifi on the dream machine itself" are you running home assistant on udm pro?

edwardmp:
Yes, exactly

Kane610:
Can't you get home assistant to talk to the external address of the udm pro? I mean that's how its intended to work, your're bypassing part of what is unifi os.

edwardmp:
I haven't tried that yet. But would this imply I need to open port 443 for this?

Otherwise, with the before mentioned patch, everything is working fine. And as I linked above another library that had the same issue, now works fine with this fix.

I can still add a test for the suggested change, just let me know which direction you wan to go here

Kane610:
The change is more or less the same right? So its perfectly fine to continue as is. A proper test is required though since I've run into issue refactoring stuff and not catching all cases. Learned my lesson :)

aiounifi/controller.py
```diff
@@ -92,7 +92,9 @@ async def check_unifi_os(self) -> None:
         response = await self._request("get", url=self.url, allow_redirects=False)
         if response.status == 200:
             self.is_unifi_os = True
-            self.headers = {"x-csrf-token": response.headers.get("x-csrf-token")}
+            # the header isn't always set, prevent it being set to none, as it would cause a NPE when serializing the headers elsewhere 
+            if response.headers.get("x-csrf-token") is not None:
+                self.headers = {"x-csrf-token": response.headers.get("x-csrf-token")}
```
Kane610, line 97:
```suggestion
            if (x_csrf_token := response.headers.get("x-csrf-token")) is not None:
                self.headers = {"x-csrf-token": x_csrf_token}
```

Kane610:
Closing this due to inactivity

PR #94, opened 2021-11-28
Add support for wlanconf:sync

dennisdegreef:
Hi :)

I want to be able to process `wlanconf:sync` messages, so that I can make switches out of the wlanconf's in HomeAssistant in order to turn the wifi on and off.

Please let me know if anything else is needed, and if I'm on the right track.

Thanks!


Kane610 (changes requested):
Thanks! This change require adding tests as well.

IIRC I initially had a working implementation for controlling availability of WLANs a couple of years ago, but back then the whole access point would disappear while changing state of a WLAN. Which isn't a really nice experience, consider this behaviour as well when progressing.

PR #98, opened 2022-01-09
Add support for outlet_overrides for plug devices

cometmth:
Ultimately I'd like to be able to control UniFi outlets from Home Assistant.  Starting here by adding support for the outlet_overrides device property and a function to set the relay state of an outlet.  Tested with a single USP-Plug (model reported as UP1).


Kane610:
Thanks! I'll have a look tonight

Kane610:
Can you add a test with some real world data to show how it is represented? Maybe this would warrant a new device type rather than expanding the default device class?

cometmth:
Yeah, I'll work on writing a test for a plug device with data from my system.  

I'm not sure about making it a separate device type.  The plug shows up in the UniFi Network controller in the devices list with the other switches and access points.  It also seems to share many of the same attributes with those devices and use the same API endpoints.  

As far as I can tell, the only way to know if a device is a SmartPower device or not is to either know the list of model strings for those devices, or maybe check to see if outlet_overrides exists.

Kane610:
> I'm not sure about making it a separate device type. The plug shows up in the UniFi Network controller in the devices list with the other switches and access points. It also seems to share many of the same attributes with those devices and use the same API endpoints.
> 
> As far as I can tell, the only way to know if a device is a SmartPower device or not is to either know the list of model strings for those devices, or maybe check to see if outlet_overrides exists.

Lets start with the test and see the similarities if its worth doing its own subclass or not.

Kane610 (changes requested):
Looks good! Can you expand the test to verify calling async_set_outlet_relay_state?

Kane610:
FYI I plan to do a release of aiounifi within the next few days

cometmth:
I've added a test for that function similar to the test for the async_set_port_poe_mode.  I've also added an outlet_table property to the device class and an outlet class.  I'm not sure if that was all necessary, but I was just mimicking what was there for POE.  

I have recently acquired a a usp-strip-us, and can work on adding a test for that device as well.

Kane610:
> I have recently acquired a usp-strip-us, and can work on adding a test for that device as well.

Yes please!

cometmth:
I've added a test for the STRIP_UP6 device as well.  In the process, I noticed that I could set the json to anything at all and the assert verify_call would pass the test.  That seems to be resolved by changing 'and' to 'or' in that function.  Also note that I had to lock pytest-aiohttp==0.3.0 to get the tests to run.  It seems that version 1.0.0 and newer break the tests with the following error.

AttributeError: module 'pytest_asyncio' has no attribute 'fixture'

aiounifi/devices.py
```diff
@@ -345,3 +384,77 @@ def __getitem__(self, obj_id: int) -> Port:
...
+        """Is outlet power on."""
+        return self.raw.get("relay_state")
+
+
+class Outlets:
+    """Represents outlets on a device."""
+
+    def __init__(self, raw: List[dict]) -> None:
+        """Initialize port manager."""
+        self.outlets: Dict[Union[int, str], Port] = {}
```
Kane610, line 427: Copy paste issue
```suggestion
        self.outlets: Dict[int, Outlet] = {}
```

aiounifi/devices.py
```diff
@@ -345,3 +384,77 @@ def __getitem__(self, obj_id: int) -> Port:
...
+        for raw_outlet in raw:
+            outlet = Outlet(raw_outlet)
+
+            if (index := outlet.index) is not None:
+                self.outlets[index] = outlet
+            elif ifname := outlet.ifname:
```
Kane610, line 433: Copy paste issue, outlets don't have ifname

aiounifi/devices.py
```diff
@@ -345,3 +384,77 @@ def __getitem__(self, obj_id: int) -> Port:
...
+class Outlets:
+    """Represents outlets on a device."""
+
+    def __init__(self, raw: List[dict]) -> None:
+        """Initialize port manager."""
+        self.outlets: Dict[Union[int, str], Port] = {}
+        for raw_outlet in raw:
+            outlet = Outlet(raw_outlet)
+
+            if (index := outlet.index) is not None:
```
Kane610, line 431: Can index be None? It would be good if it could be expected to exist

aiounifi/devices.py
```diff
@@ -345,3 +384,77 @@ def __getitem__(self, obj_id: int) -> Port:
...
+
+    def update(self, raw: List[dict]) -> None:
+        """Update outlets."""
+        for raw_outlet in raw:
+            index = None
+
+            if "index" in raw_outlet:
+                index = raw_outlet["index"]
+
+            elif "ifname" in raw_outlet:
```
Kane610, line 444: Copy paste issue; Ifname is for ports

aiounifi/devices.py
```diff
@@ -345,3 +384,77 @@ def __getitem__(self, obj_id: int) -> Port:
...
+
+    @property
+    def name(self) -> str:
+        """Name of outlet"""
+        return self.raw.get("name", "")
+
+    @property
+    def index(self) -> int:
+        """Outlet index."""
+        return self.raw.get("index")
```
Kane610, line 404: Index always exist? Current typing suggests that an int will always be returned, but here None can also be returned.

I think we can expect it to exist.

If you actually expect there be situations where index will be None please add a test for it.

aiounifi/devices.py
```diff
@@ -49,6 +49,7 @@ def __init__(
         """Initialize device."""
         super().__init__(raw, request)
         self.ports = Ports(raw.get("port_table", []))
+        self.outlets = Outlets(raw.get("outlet_table", []))
```
Kane610, line 52:
```suggestion
        self.outlets = Outlets(raw.get("outlet_table", []))
```
Make sure to do the same after line 61 to make it possible to update states with the strip as well

aiounifi/devices.py
```diff
@@ -345,3 +384,77 @@ def __getitem__(self, obj_id: int) -> Port:
     def __iter__(self) -> Iterator[Union[int, str]]:
         """Iterate over ports."""
         return iter(self.ports)
+
+
+class Outlet:
```
Kane610, line 389: Can most properties of outlet be expected to exist?, then you can replace all raw.get with raw[]

aiounifi/devices.py
```diff
@@ -345,3 +384,77 @@ def __getitem__(self, obj_id: int) -> Port:
...
+
+    @property
+    def has_metering(self) -> Optional[bool]:
+        """Is metering supported."""
+        return self.raw.get("has_metering")
+
+    @property
+    def relay_state(self) -> Optional[bool]:
+        """Is outlet power on."""
+        return self.raw.get("relay_state")
```
Kane610, line 419: All examples you've shared of relay_state shows it to always exist
```suggestion
    def relay_state(self) -> bool:
        """Is outlet power on."""
        return self.raw["relay_state"]
```

tests/test_controller.py
```diff
@@ -65,7 +65,7 @@ def verify_call(aioresponse: tuple, method: str, url: str, **kwargs: dict) -> bo
             successful_match = True
 
             for key in kwargs:
-                if key not in call[1] and call[1][key] != kwargs[key]:
+                if key not in call[1] or call[1][key] != kwargs[key]:
```
Kane610, line 68: Interesting, well if all tests continue to pass.

Kane610 (changes requested):
Some changes to do, most comments can be viewed in a generic way, so they might apply to multiple places.

I will re-review tomorrow again with a bit clearer head.

aiounifi/devices.py
cometmth: From the devices I've seen so far, index always exists and in an integer.  Will fix.

aiounifi/devices.py
cometmth: I agree that index will likely always exist.  I'm a bit out of my depth here.  I see that None will be returned if self.raw.get("index") doesn't produce a value, but I'm not sure what to do about that.  I don't think it makes sense to pick a default integer to return instead.  The Port class uses Optional[int] for port_idx which then allows None to be returned.  Should I set this up that way?

aiounifi/devices.py
cometmth: I haven't seen an example of it yet, but I'd imagine that if has_relay is False, there might not be a relay_state property.  similarly, all of my outlets have has_metering set to False.  I don't see any properties about a metering status, or current or past power usage, but I would assume if the outlet had metering there would be additional properties related to metering.

aiounifi/devices.py
cometmth: Ah, I see from the example about relay_state below.  I'll remove the .get() part.

aiounifi/devices.py
cometmth: I changed name, index, has_relay, and has_metering to use self.raw[] instead of self.raw.get().  I think relay_state might be optional depending on if the outlet has a relay or not.  I also added another optional property "cycle_enabled".  cycle_enabled seems to show up by default on the UP6, but is only present on the UP1 after flipping the "Modem Power Cycle" switch in the UniFi Network controller.

tests/test_controller.py
cometmth: I ran the existing tests that used that function and they do pass, but I had to fix some mistakes in the tests I added.

cometmth:
Thank you for reviewing in such detail!

tests/test_controller.py
cometmth: I ran the rest of the tests and found two that fail after this change in test_controller.py.  test_controller() fails at line 106.
`call[1][key]: {'username': 'user', 'password': 'pass', 'remember': True}`
`kwargs[key]: {'username': 'user', 'password': 'password', 'remember': True}`

test_unifios_controller fails at line 199 and 212.
`call[1][key]: {}`
`kwargs[key]: {'x-csrf-token': '123'}`

`call[1][key]: {'username': 'user', 'password': 'pass', 'remember': True}`
`kwargs[key]: {'username': 'user', 'password': 'password', 'remember': True}`

I have not committed fixes for these tests, since I didn't write them.  Should I revert the change to verify_call() since it's not strictly required to add support for outlets and is probably out of scope for this PR?

Kane610: Yeah change it back, lets manage that separately

aiounifi/devices.py
Kane610: Lets do it strict at least initially, easier to catch if some new device change the way it reports

aiounifi/devices.py
Kane610: Ok, what does cycle_enabled do?

aiounifi/devices.py
```diff
@@ -345,3 +385,76 @@ def __getitem__(self, obj_id: int) -> Port:
...
+
+    def values(self) -> ValuesView[Outlet]:
+        """Return outlets."""
+        return self.outlets.values()
+
+    def __getitem__(self, obj_id: int) -> Outlet:
+        """Get specific port based on key."""
+        return self.outlets[obj_id]
+
+    def __iter__(self) -> Iterator[Union[int, str]]:
```
Kane610, line 458:
```suggestion
    def __iter__(self) -> Iterator[int]:
```

aiounifi/devices.py
```diff
@@ -345,3 +385,76 @@ def __getitem__(self, obj_id: int) -> Port:
...
+    def update(self, raw: List[dict]) -> None:
+        """Update outlets."""
+        for raw_outlet in raw:
+            index = None
+
+            if "index" in raw_outlet:
+                index = raw_outlet["index"]
+
+            if index in self.outlets:
+                self.outlets[index].raw = raw_outlet
```
Kane610, line 448: We're saying we expect index to exist right so lets just do this
```suggestion
            index = raw_outlet["index"]
            self.outlets[index].raw = raw_outlet
```

aiounifi/devices.py
```diff
@@ -345,3 +385,76 @@ def __getitem__(self, obj_id: int) -> Port:
...
+
+    def values(self) -> ValuesView[Outlet]:
+        """Return outlets."""
+        return self.outlets.values()
+
+    def __getitem__(self, obj_id: int) -> Outlet:
+        """Get specific port based on key."""
```
Kane610, line 455:
```suggestion
        """Get specific outlet based on key."""
```

aiounifi/devices.py
```diff
@@ -200,7 +212,35 @@ def wlan_overrides(self) -> list:
         """Wlan configuration override."""
         return self.raw.get("wlan_overrides", [])
 
-    async def async_set_port_poe_mode(self, port_idx: int, mode: str) -> List[dict]:
+    async def async_set_outlet_relay_state(self, outlet_idx: int, state: bool) -> list[dict]:
+        """Set outlet relay state.
+
+        true:  outlet power output on
+        false: outlet power output off
```
Kane610, line 219:
```suggestion
        True:  outlet power output on.
        False: outlet power output off.
```

Kane610 (changes requested):
Nearly there :)
You will also need to do a rebase so you're inline with the other changes of devices.py

aiounifi/devices.py
cometmth: It's used to indicate which outlet(s) your modem or other internet equipment is connected to.  If you have a UniFi Security Gateway, or a Dream Machine, or some other UniFi router, the router will inform the network controller when the internet goes down.  If you have one or more outlets with cycle_enabled=True, the UniFi network controller will power cycle those outlets to try and regain internet connectivity.  For the purposes of this library, it's not super important and is really just informational.  Maybe someone wants to use that flag to have hass restart their modem under certain conditions?

Kane610:
Quick glance on mobile looks good! Will review it on computer shortly

aiounifi/devices.py
```diff
@@ -338,3 +378,70 @@ def __getitem__(self, obj_id: int) -> Port:
...
+
+    @property
+    def relay_state(self) -> bool:
+        """Is outlet power on."""
+        return self.raw["relay_state"]
+
+    @property
+    def cycle_enabled(self) -> bool | None:
+        """Modem Power Cycle"""
+        return self.raw.get("cycle_enabled")
```
Kane610, line 418: Last thing! It seems in all examples in the tests the cycle is either True or False, should this be more strict as well?

Either to return False if key does not exist or always expect the value to exist

aiounifi/devices.py
Kane610: As a separate PR, could you look into how to set cycle_enabled True/False resulting in a method to do that?

Kane610:
Looks good! Just that final thing with cycle enabled to take a stance on

aiounifi/devices.py
cometmth: In the PLUG_UP1 test it is set to `assert outlet_1.cycle_enabled is None`.  That device does not report that property out of the box.  It is probably as safe bet that if cycle_enabled is missing it can be assumed to be False.  I'll make that change.

Kane610: Sorry missed that, yes keeping it to only bool simplifies things with regards to typing

Kane610 (changes requested):
Fixed your flake8 comments from test run

Kane610:
I'm doing some minor fixes and code coverage improvements to get this done, I will ignore the type issue for headers as that is irrelevant to this PR.

aiounifi/devices.py
cometmth: I'd be happy to.

Kane610:
Awesome! thanks @cometmth !

Kane610:
@cometmth would you mind looking at how to set cycle_enabled/disabled and adding that in a separate PR as well?

Kane610:
@cometmth PR up to add outlet support to UniFi integration https://github.com/home-assistant/core/pull/64859

cometmth:
Thanks for taking this to the finish line and writing the home assistant integration!

Kane610:
Thanks for your contribution @cometmth !

PR has been merged 🎉 

Could you run home assistant dev or beta so we flush out any possible issues?

cometmth:
@Kane610 I'm currently running the dev20220126 build, and it seems to be mostly working!  Both outlet capable devices were discovered immediately and added in to HA.  Only two small issues.  

1. The UP6 in my UniFi Network Controller is named "USP-Strip", but in HA the device shows up as "UP6 Outlet 1".  That device does actually have switches for all 7 outlets on the device, but the name seems wrong.  I have not looked into why yet.  The UP1 device is correctly named "Plug" in both my UniFi Network controller and in HA.  

2. Probably nothing can be done for this, but just noticed that because UniFi takes so long (1-2s) to actually change the state of the outlets, the switch in HA goes back and forth a bit.  For example, if an outlet is on and I turn it off in HA, the toggle goes to "off" when I click it, and then about a second later goes back to on.  When the outlet finally does turn off, the toggle correctly reflects that and moves to the off position.  If that's just cosmetic, I think it's fine, but if that has the potential to impact some automations, I'd wonder if we could give more time for the controller to respond to reduce the "jitter".

Kane610:
> 

1. Where does the name originate, I looked at your fixture and there is neither name nor hostname. Right now the name is defined like this `self._attr_name = f"{device.name or device.model} {device.outlets[index].name}"`
2. What you can do is to enable debug logging in the integration, if you can see any events related to the outlets we can utilise them for quicker signalling. The other way would be to assume that the state changes with the call.

kerhbal:
I guess this is a better place to discuss the test on this new feature. I'm coming from this topic: https://github.com/home-assistant/home-assistant.io/issues/20814
So I've started a separate home assistant dev instance for testing, and added unifi integration, but something is not right here, probably I'm doing it wrong.

- My stable version: core-2021.12.9 
- My test version: core-2022.3.0.dev20220127
- This is the integration from my stable version: 
Vivaldi-Configuration - Home Assistant - Vivaldi--2022-01-28 at 12 15 25

- This is the configure popup for my stable version: 



- This is the integration from my test version:
Vivaldi-Configuration – Home Assistant - Vivaldi--2022-01-28 at 12 16 12

- This is the configure popup for my test version: 
Vivaldi-Configuration – Home Assistant - Vivaldi--2022-01-28 at 12 20 17

And the entities in test version are just a bunch of device trackers, no device at all. I used same unifi local credential for the integration. same network.

@cometmth @Kane610 any ideas? Thank you!

Kane610:
I guess that your home assistant user in unifi does not have administrator controls?

kerhbal:
@Kane610 oh shoot, I probably used a different user on my stable version. yeah I just changed permission for the user being used by test version. it works! Since I only have a plug, just tested on it.
- switch toggle in hass will reflect on the actual device
- toggle on the actual device also will reflect in hass
- delay is minimal, less than 0.5s
Thank you!

Kane610:
Sweet!

PR #111, opened 2022-01-30
Add set_outlet_cycle_enabled() method.

cometmth:
As requested, a method to set cycle_enabled for an outlet.  This is set with outlet_overrides in the same way relay_state is set.


Kane610:
Sweet! I'll have a look tonight

Kane610 (approved):
Great! Thanks!

PR #116, opened 2022-03-19
Use x-csrf-token only when remote site provides it.

arekm:
Don't fail with 'TypeError: Cannot serialize non-str key None' when
remote site doesn't provide x-csrf-token. That can happen if we are
connecting to invalid site (like port 443 on unifi controller instead
of 8443).


aiounifi/controller.py
```diff
@@ -111,7 +111,9 @@ async def check_unifi_os(self) -> None:
             response := self.last_response
         ) is not None and response.status == HTTPStatus.OK:
             self.is_unifi_os = True
-            self.headers = {"x-csrf-token": response.headers.get("x-csrf-token")}
+            csrf_token = response.headers.get("x-csrf-token")
+            if csrf_token is not None:
+                self.headers = {"x-csrf-token": csrf_token }
```
Kane610, line 116:
```suggestion
            if (csrf_token := response.headers.get("x-csrf-token")) is not None:
                self.headers = {"x-csrf-token": csrf_token }
```

Kane610 (changes requested):
Thank you for contributing.

You need to also add tests to cover this new behaviour. I have previously experienced a lot of extra work when reworking areas of the code that's been contributed by others that don't come with tests.

Alienero:
I have same issue, whe I using UDM-PRO. https://ubnt doesn't provide csrf token

PR #134, opened 2022-06-03
Improve events

Kane610:


aiounifi/interfaces/events.py
```diff
@@ -0,0 +1,124 @@
...
+import logging
+from typing import Any, Callable, Final, Optional
+
+from ..models.event import EventKey, MessageKey, WebsocketData
+
+LOGGER = logging.getLogger(__name__)
+
+
+SubscriptionType = tuple[
+    Callable[[WebsocketData], set[str]],
```
J3173, line 14: This type should include Event:
```suggestion
    Callable[[WebsocketData], set[str | Event]],
```
to match with https://github.com/Kane610/aiounifi/blob/21b9018427155d2c9b339c4e2125f3d2409f1c42/aiounifi/interfaces/api.py#L59

aiounifi/models/event.py
```diff
@@ -305,3 +306,39 @@ def version_from(self) -> str:
...
+    event: Event
+    data: dict[str, Any]
+
+    @classmethod
+    def from_dict(cls, data: dict[str, Any]) -> "WebsocketData":
+        """Create data container instance from dict."""
+        return cls(
+            meta=Meta.from_dict(data.get("meta", "")),
+            event=Event(data["data"]),
+            data=data,
```
J3173, line 343: I would only put the data (without meta) in the data field. It is the part that is used from the data field and meta is separately available.
```suggestion
            data=data["data"],
```

aiounifi/interfaces/api.py
```diff
@@ -22,66 +22,84 @@ class APIItems:
...
+        if data.meta.message in self.process_messages:
+            return self.process_item(data.data["data"])
 
-            for callback in self._subscribers:
-                callback("added", key)
+        if data.meta.message in self.remove_messages:
+            return self.remove_item(data.data["data"])
 
-        return new_items
+        return ""
```
J3173, line 70: I would have used None as return value here and in the three methods below to indicate nothing is returned.
```suggestion
        return None
```

J3173 (commented):
I like moving the message_handler to a separate handler class.

However in this PR the terms 'message' and 'event' seem to be used in interchangeably and that is confusing to me.

Maybe you can consider making a separate MessageHandler and EventHandler? Then the EventHandler can use the MessageHandler to subscribe to messages with MessageKey.EVENT.

aiounifi/interfaces/api.py
Kane610: This is just to support current functionality, this will be completely replaced by the subscribe functionality, but I don't want to break things too early on.

Kane610:
> I like moving the message_handler to a separate handler class.
> 
> However in this PR the terms 'message' and 'event' seem to be used in interchangeably and that is confusing to me.
> 
> Maybe you can consider making a separate MessageHandler and EventHandler? Then the EventHandler can use the MessageHandler to subscribe to messages with MessageKey.EVENT.

This was my primary concern but went towards the simpler implementation. I initially had them separate but still in the same event class. You think messages and events should be separate classes?

Unfortunately I will be away for the next 10 days so can't progress on unifi until after that. Sense of guilt not finishing this to move your PRs along, sorry!

Kane610:
@J3173 thanks a lot, it was a huge improvement separating messages and events. Just gonna look over it tomorrow again and then test it out with the integration and I can merge it.

Apologize it took me so long to revisit this, after this we can get your PRs moving again.

PR #137, opened 2022-06-11
Add device upgrade command

J3173:
This allows the firmware of UniFi network devices to be upgraded by calling `unifi_controller.devices.upgrade(mac="80:2a:a8:00:01:02")`.

### DIFF
--- a/aiounifi/interfaces/devices.py
+++ b/aiounifi/interfaces/devices.py
@@ -12,6 +12,8 @@ from .api import APIItems
 
 URL: Final = "/stat/device"
 
+URL_DEVICE_MANAGER: Final = "/cmd/devmgr"
+
 
 class Devices(APIItems):
     """Represents network devices."""
@@ -19,3 +21,8 @@ class Devices(APIItems):
     KEY = "mac"
     path = URL
     item_cls = Device
+
+    async def upgrade(self, mac: str) -> list[dict]:
+        """Upgrade network device."""
+        data = {"mac": mac, "cmd": "upgrade"}
+        return await self.controller.request("post", URL_DEVICE_MANAGER, json=data)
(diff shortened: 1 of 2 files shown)


Kane610:
Thanks! I will have a look shortly.

Kane610 (commented):
This is nice! Thank you!

Are you ready to publish a PR to add this to HASS as well?

I would prefer if the method is a bit simplified and future proof (that's what Im doing all over the library right now)

Maybe change the method to be `device_manager_control` or something and the command itself would be a enum instead, what do you think?

J3173:
> Are you ready to publish a PR to add this to HASS as well?

I have done an initial implementation and tested it manually. But I still need to write some tests to make it a full PR.

> I would prefer if the method is a bit more future proof (that's what Im doing all over the library right now)
> 
> Maybe change the method to be `device_manager_control` or something and the command itself would be a enum instead, what do you think?

That could be possible, but when I look at https://github.com/Art-of-WiFi/UniFi-API-client/blob/master/src/Client.php the (number of) parameters are not the same for the different commands. You could add a dict for the parameters, but I think that separate methods would be clearer for the user.

Kane610:
> I have done an initial implementation and tested it manually. But I still need to write some tests to make it a full PR.

Great!

> That could be possible, but when I look at https://github.com/Art-of-WiFi/UniFi-API-client/blob/master/src/Client.php the (number of) parameters are not the same for the different commands. You could add a dict for the parameters, but I think that separate methods would be clearer for the user.

Your implementation differs from what's inside the example; https://github.com/Art-of-WiFi/UniFi-API-client/blob/cbe89d913cc4dd3742d5fd35e201ec86bc8ccdc9/src/Client.php#L2858

J3173:
I'm working on the Home Assistant PR at https://github.com/J3173/home-assistant-core/tree/unifi-update-install

> The url differs from what's inside the example; https://github.com/Art-of-WiFi/UniFi-API-client/blob/cbe89d913cc4dd3742d5fd35e201ec86bc8ccdc9/src/Client.php#L2858
> 
> Does this work as well?

I've used the url which the UniFi web interface also uses. Both urls seem to work.

PR #39, opened 2020-07-02
Retry login when session expires

bdraco:
Verified replication and fix for https://github.com/home-assistant/core/issues/33024

If we need tests I probably need to bring in `aioresponses` to mock specific urls

PR #140, opened 2022-06-16
use x-csrf-token from login on UDM platform

stonith:
The x-csrf-token from the login response should be used instead of the token from the base url response.

fixes https://github.com/home-assistant/core/issues/73071 and https://github.com/home-assistant/core/issues/71017

### DIFF
--- a/aiounifi/controller.py
+++ b/aiounifi/controller.py
@@ -113,8 +113,6 @@ class Controller:
             response := self.last_response
         ) is not None and response.status == HTTPStatus.OK:
             self.is_unifi_os = True
-            if (csrf_token := response.headers.get("x-csrf-token")) is not None:
-                self.headers = {"x-csrf-token": csrf_token}
 
     async def login(self) -> None:
         """Log in to controller."""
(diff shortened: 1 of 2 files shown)


Kane610:
Thanks! I'll have a look tonight

Kane610:
You can add a small test that can verify the expected behavior

aiounifi/controller.py
```diff
@@ -130,7 +128,11 @@ async def login(self) -> None:
         }
 
         await self._request("post", url=url, json=auth)
-
+        if (
+            response := self.last_response
+        ) is not None and response.status == HTTPStatus.OK:
+            if (csrf_token := response.headers.get("x-csrf-token")) is not None:
+                self.headers = {"x-csrf-token": csrf_token}
```
Kane610, line 135:
```suggestion
        if (
            (response := self.last_response) is not None
            and response.status == HTTPStatus.OK
            and (csrf_token := response.headers.get("x-csrf-token")) is not None
        ):
            self.headers = {"x-csrf-token": csrf_token}
```

No need to nest if-statement

Kane610 (changes requested):
I've tested your change and it works fine! Thanks!

Fix failing tests and my comment and I can put out a new release.

stonith:
I'll try and fix tests tonight.

stonith:
I don't quite follow what test_unifios_controller_relogin_success test is doing, is this needed?

Kane610:
> I don't quite follow what test_unifios_controller_relogin_success test is doing, is this needed?

https://github.com/Kane610/aiounifi/pull/39

Kane610:
Looks good! I will look through it tonight and build a new release if everything is as it looks!

Kane610 (approved):

PR #141, opened 2022-06-18
Support incremental device information updates

J3173:
`device:sync` messages are only send out when a UniFi device is in the _Online_ state.

This implements the `device:update` message so device information is updated when a device is in another state, e.g. _Updating_. This is needed to display up-to-date state of the device in Home Assistant when the device is updating.

The data in `device:update` contains only the fields that are modified since the previous `device:sync` or `device:update` message. This PR adds a new `incremental_update` method to the APIItem class which updates the exiting `raw` data with the newly received information.

This PR also modifies existing tests to deepcopy the fixture data when calling `process_raw` or setting `websocket._data` to prevent accidental modifying the fixture data.

Tested with UniFi Network Application version 7.1.66.

### DIFF
--- a/aiounifi/controller.py
+++ b/aiounifi/controller.py
@@ -40,6 +40,7 @@ LOGGER = logging.getLogger(__name__)
 MESSAGE_CLIENT: Final = "sta:sync"
 MESSAGE_CLIENT_REMOVED: Final = "user:delete"
 MESSAGE_DEVICE: Final = "device:sync"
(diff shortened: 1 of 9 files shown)


Kane610:
Awesome!

I'm doing some streamlining and improvements with internal signalling (https://github.com/Kane610/aiounifi/pull/134) which will break this. I'm working on these to better scale changes in the future. Im gonna try to get it done this weekend to not block you too much. Feedback is appreciated.

PR #145, opened 2022-06-23
Switch to orjson for loading json

bdraco:
Home Assistant now has support for orjson via https://github.com/home-assistant/core/pull/73849

This json loads comes up as a hotspot in profiles often (https://community.home-assistant.io/t/how-to-install-py-spy-on-a-ha-os-instance-please-instruct/430822/29?u=bdraco) and https://community.home-assistant.io/t/how-to-install-py-spy-on-a-ha-os-instance-please-instruct/430822/45?u=bdraco

Screen Shot 2022-06-22 at 21 37 42

Switching it out results in a dramatic savings and makes the json decode disappear from the `py-spy` as its no longer slow enough to even show up, and we see the smaller functions that were obscured by the json overhead

Screen Shot 2022-06-22 at 21 35 30


Kane610 (approved):
Thanks @bdraco ! I will get this merged before the beta next week!

PR #146, opened 2022-06-25
set can_retry_login to true by default

stonith:
Session expiry never retries login because can_retry_login is set to False by default.

Fixes #73071 again

### DIFF
--- a/aiounifi/controller.py
+++ b/aiounifi/controller.py
@@ -90,7 +90,7 @@ class Controller:
         self.site = site
         self.sslcontext = sslcontext
         self.callback = callback
-        self.can_retry_login = False
+        self.can_retry_login = True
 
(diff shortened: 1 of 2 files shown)


Kane610:
Damn it. Didn't think it through. The bool needs to be set in the login method. Which you removed in the previous PR. Revert the bool to false in init and set it to true at the end of the login method instead.

stonith:
> Damn it. Didn't think it through. The bool needs to be set in the login method. Which you removed in the previous PR. Revert the bool to false in init and set it to true at the end of the login method instead.

Ah damn, I didn't notice I removed that! I was working off 0.31 which was slightly different and didn't test master with HA so didn't catch it, sorry.

Kane610:
It got varggrin in the tight moment at least

PR #147, opened 2022-06-26
move can_retry_login back to login function

stonith:
As per title, this is needed because after single retry can_retry_login isn't reset to true whereas if login is completed again it will get reset.


Kane610:
I'm back home in the afternoon. Then I can do a new release

PR #152, opened 2022-07-02
Add some additional properties to `Device`

joshuaspence:
Supporting change for home-assistant/core#74025.

### DIFF
--- a/aiounifi/models/device.py
+++ b/aiounifi/models/device.py
@@ -52,6 +52,24 @@ class Device(APIItem):
         return self.raw["considered_lost_at"]
 
     @property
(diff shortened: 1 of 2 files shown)


Kane610 (commented):
Could we create two data classes for temperatures and system-stats and handle them similar to ports and outlets (only simpler)

Kane610:
Any plans to continue this @joshuaspence ?

joshuaspence:
Sorry, life got in the way

Kane610:
> Sorry, life got in the way

It happens! Thanks for replying.

PR #206, opened 2022-10-07
Add EVT_XG_OutletPowerCycle event key

ahutchings:
I am seeing a " Unsupported event key EVT_XG_OutletPowerCycle" log message in Home Assistant. This event is triggered when the UXG Pro cycles its SmartPower outlet (intended for use with an external modem).


Kane610:
Would you mind share a printout from debug logs with the full event for future reference?

I will package and publish a new release of aiounifi for home assistant on Sunday.

ahutchings:
Thanks! The event isn't getting fired when I manually cycle outlet power, so I may need to pull the plug on the modem to trigger it. I'll try that this evening.

PR #391, opened 2023-06-23
Add port support to message class

jasperslits:
This adds port forward to the Message class to avoid a warning when there is Port forwarding settings in the Unifi Network application are added or updated.


Kane610:
Awesome! Thank you!

PR #397, opened 2023-07-03
Add message "unifi-device:add"

Kane610:
https://github.com/home-assistant/core/issues/95652


tankdeer:
This did not resolve the issue. Still exists after updating to 2023.7.

I think the issue might be that the this PR added the message `unifi-device:add`, but the message in the error is just `device:add`

PR #409, opened 2023-07-17
Add traffic rules to library

herriejr:
With these changes, it is possible to control the trafficrules, defined in the unifi network management.
At this moment you can enable and disable a selected traffic rule.
(pls bear with me, since this is my 1st pull request)


Kane610:
Hello! Thank you for your PR, I started doing a review but got lost in other stuff, I will get back to you shortly!

aiounifi/models/trafficrule.py
```diff
@@ -0,0 +1,123 @@
...
+        """Create traffic rule request."""
+        return cls(method="get", path="/trafficrules", data=None)
+
+@dataclass
+class TrafficRuleEnableRequest(ApiRequest):
+    """Request object for wlan enable."""
+
+    v2: bool = True
+    @classmethod
+    def create(cls, trafficrule: dict[str, Any], enable: bool) -> "TrafficRuleEnableRequest":
```
Kane610, line 79:
```suggestion
    def create(cls, traffic_rule: TypedTrafficRule, enable: bool) -> "TrafficRuleEnableRequest":
```

Kane610 (changes requested):
Its a good start, but some improvements are needed and of course tests.

herriejr:
I was working on some other projects and was on holiday. I will start implementing the requested changes and start adding the tests.

Kane610:
I recommend starting with doing a rebase as I've refactored some aspects lg the library

herriejr:
Ok, I started with a rebase and I'm not completely done yet (still some cleaning up to do of the code and I still need to squeeze an update in the code). But the rebase solved the problem for which I added the 'v2-bool' in the first place.
I think I did something wrong, since I accidentally closed this pull request :innocent:. I will try to do the missing modifications in the next few days and add the tests in the following weeks.
I will also create a new pull request in the next few days.

Kane610:
Take your time!

herriejr:
I made the requested modifications, and based it on the new structure.
I had to change a line in the api_handlers.py-file, since the response of this API is different (does not contain 'meta' nor 'data'-keys). I also added a few tests for the traffic rules to the tests.

aiounifi/models/traffic_rule.py
```diff
@@ -0,0 +1,134 @@
+"""Traffic rules as part of a UniFi network."""
+
+from dataclasses import dataclass
+
+from typing_extensions import NotRequired, TypedDict
```
Kane610, line 5: The library expects python3.11 so you can import them from typing
```suggestion
from typing import NotRequired, TypedDict
```

aiounifi/models/traffic_rule.py
```diff
@@ -0,0 +1,134 @@
...
+
+    def full_path(self, site: str, is_unifi_os: bool) -> str:
+        return f"/proxy/network/v2/api/site/{site}{self.path}"
+
+@dataclass
+class TrafficRuleListRequest(TrafficRuleRequest):
+    """Request object for traffic rule list."""
+
+    @classmethod
+    def create(cls) -> "TrafficRuleListRequest":
```
Kane610, line 81: With python 3.11 you can import Self from typing
```suggestion
    def create(cls) -> Self:
```

tests/conftest.py
```diff
@@ -158,6 +158,26 @@ def mock_get_request(path: str, unifi_path: str, payload: dict[str, Any]) -> Non
...
+
+    def mock_get_request(path: str, unifi_path: str, payload: dict[str, Any]) -> None:
+        """Register HTTP response mock."""
+        url = unifi_path if is_unifi_os else path
+        data = payload
+        mock_aioresponse.get(f"https://host:8443{url}", payload=data)
+
+    mock_get_request(
+        "/proxy/network/v2/api/site/default/trafficrules",
+        "/proxy/network/v2/api/site/default/trafficrules",
```
Kane610, line 177: One should be for devices running unifi os and one should be for self hosted, Im guessing its just removing /proxy/network
```suggestion
        "/v2/api/site/default/trafficrules",
        "/proxy/network/v2/api/site/default/trafficrules",
```

tests/conftest.py
```diff
@@ -158,6 +158,26 @@ def mock_get_request(path: str, unifi_path: str, payload: dict[str, Any]) -> Non
...
+    is_unifi_os: bool,
+    traffic_rule_payload: dict[str, Any],
+) -> None:
+    """Use fixtures to mock traffic rule endpoint."""
+
+    def mock_get_request(path: str, unifi_path: str, payload: dict[str, Any]) -> None:
+        """Register HTTP response mock."""
+        url = unifi_path if is_unifi_os else path
+        data = payload
+        mock_aioresponse.get(f"https://host:8443{url}", payload=data)
```
Kane610, line 173: This should be moved into the real endpoint_fixture

tests/test_traffic_rules.py
```diff
@@ -0,0 +1,94 @@
...
+        json=traffic_rule_raw,
+    )
+
+    traffic_rule_raw["enabled"] = False
+    await traffic_rules.disable(traffic_rule)
+    assert unifi_called_with(
+        "put",
+        "/proxy/network/v2/api/site/default/trafficrules/6452cd9b859d5b11aa002ea1",
+        json=traffic_rule_raw,
+    )
```
Kane610, line 94: This is tested by a separate test above.

Kane610 (changes requested):
You would be helped setting up the dev env in a python3.11 venv and using the pre-commit tool. You can run the setup.sh script to do that, there are a lot of small things which would be caught by that.

Kane610:
> I made the requested modifications, and based it on the new structure.
> I had to change a line in the api_handlers.py-file, since the response of this API is different (does not contain 'meta' nor 'data'-keys). I also added a few tests for the traffic rules to the tests.

Yes we need to figure out a way to specify different expected data structures from different api endpoints.

aiounifi/controller.py
```diff
@@ -52,6 +53,7 @@ def __init__(self, config: Configuration) -> None:
         self.sites = Sites(self)
         self.system_information = SystemInformationHandler(self)
         self.wlans = Wlans(self)
+        self.traffic_rules = TrafficRules(self)
```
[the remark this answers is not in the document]
herriejr: Already made this adjustmenst locally. Thanks for the feedback.

aiounifi/controller.py
```diff
@@ -63,6 +65,7 @@ def __init__(self, config: Configuration) -> None:
             self.sites.update,
             self.system_information.update,
             self.wlans.update,
+            self.traffic_rules.update,
```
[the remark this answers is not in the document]
herriejr: Already made this adjustments locally. Thanks for the feedback.

aiounifi/models/traffic_rule.py
```diff
@@ -0,0 +1,134 @@
...
+    """Data class with required properties of a traffic rule API request."""
+    '''We need a way to indicate if, for our model, the v2 API must be called.
+    Therefore an intermediate dataclass 'TrafficRuleRequest' is made,
+    for passing the correct path. This way, we do not need to alter any of the
+    other classes that not need to know about the version of the api used.
+    With the refactoring of the aiounifi-library, this is now possible.
+    '''
+
+    def full_path(self, site: str, is_unifi_os: bool) -> str:
+        return f"/proxy/network/v2/api/site/{site}{self.path}"
```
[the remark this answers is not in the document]
herriejr: I installed the latest version of the Unifi-network controller on my NAS and checked the URL that is used in that case. I corrected this already locally (in that case "/proxy/network" is not in the path).

tests/conftest.py
herriejr: Already made this adjustments locally. Thanks for the feedback.

aiounifi/models/traffic_rule.py
herriejr: Already made this adjustments locally. Thanks for the feedback.

aiounifi/models/traffic_rule.py
herriejr: Already made this adjustments locally. Thanks for the feedback.

aiounifi/interfaces/traffic_rules.py
```diff
@@ -0,0 +1,31 @@
...
+    item_cls = TrafficRule
+    api_request = TrafficRuleListRequest.create()
+
+    async def enable(self, traffic_rule: TrafficRule) -> list[dict[str, Any]]:
+        """Enable traffic rule defined in controller."""
+        tr = self.controller.traffic_rules.get(traffic_rule.id)
+        traffic_rule_dict = tr.raw
+        return await self.controller.request(
+            TrafficRuleEnableRequest.create(traffic_rule_dict, enable=True)
+        ) and await self.controller.traffic_rules.update()
```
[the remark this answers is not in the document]
herriejr: Well, I'm also convinced that is not the right place to trigger an update (I already removed this locally). I just can't seem to grasp how the state of a traffic rule will be updated, so that the state in the controller reflects the actual state of the traffic rule. Or is this not necessary?

I inspected the messages coming from the Unifi-controller (websocket) and I don't see any messages emitted when changing the state of the traffic rule.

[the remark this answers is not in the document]
Kane610: Do the controller only do a get on the endpoint when changing configuration?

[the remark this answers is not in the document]
herriejr: The controller does a PUT on the endpoint. The response is the a JSON with the changed configuration (without meta/data-keys, just the config of the traffic rule). I have been monitoring the traffic for a while (+/- 30 minutes, also on the websocket) and I have not seen any other traffic passing concerning the PUT on the endpoint when changing the state.

tests/test_traffic_rules.py
herriejr: Indeed, I made the necessary changes.

aiounifi/interfaces/traffic_rules.py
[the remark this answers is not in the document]
Kane610: > The controller does a PUT on the endpoint. The response is the a JSON with the changed configuration (without meta/data-keys, just the config of the traffic rule). I have been monitoring the traffic for a while (+/- 30 minutes, also on the websocket) and I have not seen any other traffic passing concerning the PUT on the endpoint when changing the state.

Maybe we should pass the response back into update in the handler if you get the updated response? On the TrafficRule handler class. Consolidate enable disable methods to one to avoid the duplication and I think that could work well.




> I assume there is a slight problem in [api_handlers.py](https://github.com/Kane610/aiounifi/blob/master/aiounifi/interfaces/api_handlers.py#L99) the trafficrules-api in this line:
> 
> `self.process_raw(raw.get("data", []))`
> 
> This assumes that raw is a dict. But since the answer of the trafficrules-api is an array (of JSON-objects), the orjson.loads-function converts this [array to a list](https://docs.python.org/3/library/json.html#json-to-py-table) - I believe they also follow this conversion table.
> 
> Maybe we need to see how this affects the library and how I can best work around this.

Im considering moving initial response data parsing to the response object, could host both error parsing and preparing response for the handler.

Kane610:
Good progress! I will be away a few days now so gonna be hard to get this in for 2023.9 release, but only a few things left to resolve

herriejr:
Did it again, mistakenly closed the PR

aiounifi/interfaces/traffic_rules.py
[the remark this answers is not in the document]
herriejr: - for the data parsing of the response object, I've made a suggestion by adding a 'prepare_data' method to the ApiRequest-class (not yet completely happy with this, but it is a start). And added a call in the update()-method of the ApiHandler
- I have consolidated the enable/disable-methods (they still exist, but they call a new method 'toggle'), so code duplication is avoided. Still have to figure out a way to use the response to make the update. I'm now calling the update()-method of the handler in toggle, but using the response would be way more efficient.

aiounifi/interfaces/traffic_rules.py
```diff
@@ -0,0 +1,31 @@
...
+
+    obj_id_key = "_id"
+    path = "/trafficrules"
+    item_cls = TrafficRule
+    api_request = TrafficRuleListRequest.create()
+
+    async def enable(self, traffic_rule: TrafficRule) -> list[dict[str, Any]]:
+        """Enable traffic rule defined in controller."""
+        tr = self.controller.traffic_rules.get(traffic_rule.id)
+        traffic_rule_dict = tr.raw
```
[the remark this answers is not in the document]
herriejr: I changed the code a bit, but I didn't understand your remark 'Just like wlan'.

tests/conftest.py
herriejr: Still on my todo-list

Kane610 (commented):
Just some small comments to move you along :)

tests/conftest.py
herriejr: I get an error when adding this to _mock_endpoints when running the tests.
`ERROR tests/test_traffic_rules.py - Failed: In test_traffic_rules: function uses no argument 'traffic_rule_payload'`
I still need to investigate this further.

aiounifi/interfaces/traffic_rules.py
```diff
@@ -0,0 +1,33 @@
...
+        """Disable traffic rule defined in controller."""
+        return await self.toggle(traffic_rule, state = False)
+
+    async def toggle(self, traffic_rule: TrafficRule, state: bool):
+        """ Set traffic rule - defined in controller - to the desired state """
+        traffic_rule_dict = deepcopy(traffic_rule.raw)
+        traffic_rule_enabled = await self.controller.request(
+            TrafficRuleEnableRequest.create(traffic_rule_dict, enable=state)
+        )
+        await self.controller.traffic_rules.update()
```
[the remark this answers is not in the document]
herriejr: Has been changed, but I first call prepare_data before calling process_raw.
In this case the API just returns a single dict (a dict of the changed traffic rule) and not a list of dict(s) (as expected by process_raw). But now I'm struggling with the data-types (as reported by mypy - I started with a fresh project with a fresh .venv since I had some problems with my old .venv).

aiounifi/models/api.py
```diff
@@ -27,6 +27,8 @@ def full_path(self, site: str, is_unifi_os: bool) -> str:
             return f"/proxy/network/api/s/{site}{self.path}"
         return f"/api/s/{site}{self.path}"
 
+    def prepare_data(self, raw: TypedApiResponse):
+        return raw
```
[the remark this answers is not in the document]
herriejr: Changed.

herriejr:
I hope with the latest commits I am almost there

herriejr:
I've made some modifications to the prepare_data method I introduced. 
I've given that method some more responsibility: 
- parse the received response with orjson
- inspect the result and depending on the structure of the result, put the data in the correct way in a TypedApiResponse-object to use throughout the library

Maybe it is more logical to place this elsewhere, but now the complexity of parsing different response structures is already concentrated in one place.
I still have to solve the problem I'm having when adding traffic_rule_payload to _mock_endpoints when running the tests; since I keep getting this error:
'ERROR tests/test_traffic_rules.py - Failed: In test_traffic_rules: function uses no argument 'traffic_rule_payload'

aiounifi/models/api.py
```diff
@@ -27,6 +29,22 @@ def full_path(self, site: str, is_unifi_os: bool) -> str:
...
+        if isinstance(json_data, dict):
+            if "meta" in json_data:
+                return_data["meta"] = json_data.get("meta", {})
+                return_data["data"] = json_data.get("data", [])
+            else:
+                return_data["data"] = [json_data]
+        if isinstance(json_data, list):
+            return_data["data"] = json_data
+
+        return return_data
```
Kane610, line 46: I think the default behaviour should be the old one. Then you should override this method within `class TrafficRuleRequest(ApiRequest):` When we get more v2 apis that follow the same structure then a ApiRequestV2 could be provided as a base class.

tests/conftest.py
```diff
@@ -101,6 +101,7 @@ def endpoint_fixture(
     port_forward_payload: list[dict[str, Any]],
     site_payload: list[dict[str, Any]],
     system_information_payload: list[dict[str, Any]],
+    traffic_rule_payload: list[dict[str, Any]],
```
Kane610, line 104: The error message is something related to the two fixtures. One reason could be that both methods are defined as `def endpoint_fixture`. I still want you to make traffic_rule_payload work with the original endpoint_fixture and remove _mock_traffic_rule_endpoint

Kane610 (changes requested):

Kane610:
I only had a few minutes tonight, will look more tomorrow.

aiounifi/models/api.py
herriejr: Ok, I have moved everything that is special for the v2 api down to the TrafficRuleRequest(ApiRequest). Some little changes were still needed in other parts for working with a TypedApiResponse. I'm still finetuning some things before committing.

tests/conftest.py
herriejr: I've fixed this and made some other changes to some tests (they were not yet working with TypedApiResponse but with raw dicts). But now I have one of my test failing :confounded:
Almost there...

herriejr: I've got everything up and running. Had to make some modifications to some other tests, since process_raw now processes a TypedApiResponse (and no longer a 'raw' dict).

aiounifi/models/api.py
herriejr: Some extra conditions seemed to be required (the tests pointed me in that direction). They are added in the code.

aiounifi/interfaces/api_handlers.py
```diff
@@ -98,14 +98,15 @@ def __init__(self, controller: "Controller") -> None:
...
 
     @final
-    def process_raw(self, raw: list[dict[str, Any]]) -> None:
+    def process_raw(self, raw: TypedApiResponse) -> None:
         """Process full raw response."""
-        for raw_item in raw:
-            self.process_item(raw_item)
+        if "data" in raw:
+            for raw_item in raw["data"]:
+                self.process_item(raw_item)
```
Kane610, line 109: process_raw should not handle variations, that's why its marked `@final` lets make sure that data us prepared to an expected format.

aiounifi/interfaces/connectivity.py
```diff
@@ -156,6 +156,6 @@ async def _request(
 
 def _raise_on_error(data: "TypedApiResponse") -> None:
     """Check response for error message."""
-    if "meta" in data and data["meta"]["rc"] == "error":
+    if "meta" in data and data["meta"] is not None and data["meta"]["rc"] == "error":
         LOGGER.error(data)
         raise_error(data["meta"]["msg"])
```
Kane610, line 161: I was thinking about that method the other day, do you have any examples of errors from the v2 api? We should structure this in a scalable way.

aiounifi/interfaces/traffic_rules.py
```diff
@@ -0,0 +1,37 @@
+"""WLANs as part of a UniFi network."""
```
Kane610, line 1:
```suggestion
"""Traffic rules as part of a UniFi network."""
```

aiounifi/models/api.py
```diff
@@ -27,6 +29,16 @@ def full_path(self, site: str, is_unifi_os: bool) -> str:
...
 
+    def prepare_data(self, raw: bytes) -> TypedApiResponse:
+        """Put data, received from the unifi controller, into a TypedApiResponse."""
+        json_data = orjson.loads(raw)
+        return_data: TypedApiResponse = {}
+
+        return_data["meta"] = json_data.get("meta", None)
+        return_data["data"] = json_data.get("data", None)
+
+        return return_data
```
Kane610, line 40: Lets consider this method only supporting v1 responses and do an override in the traffic rules api request child
```suggestion
    def prepare_data(self, raw: bytes) -> TypedApiResponse:
        """Put data, received from the unifi controller, into a TypedApiResponse."""
        data: TypedApiResponse = orjson.loads(raw)
        return data
```

aiounifi/models/traffic_rule.py
```diff
@@ -0,0 +1,171 @@
...
+
+    def prepare_data(self, raw: bytes) -> TypedApiResponse:
+        """Put data, received from the unifi controller, into a TypedApiResponse."""
+        json_data = orjson.loads(raw)
+        return_data: TypedApiResponse = {}
+
+        if isinstance(json_data, dict):
+            return_data["data"] = [json_data]
+        if isinstance(json_data, list):
+            return_data["data"] = json_data
```
Kane610, line 108: Why can this be either list or dict?

tests/conftest.py
```diff
@@ -26,6 +26,12 @@ def is_unifi_os_fixture() -> bool:
     return False
 
 
+@pytest.fixture(name="v2")
+def v2_fixture() -> bool:
+    """If use UniFi OS response."""
+    return False
```
Kane610, line 32: This shouldn't be needed. You can isolate all logic to within the endpoint_fixture, either you can look at the path to see if it contains v2 or pass in an extra property to mock_get_request which says v1 = True or something

Kane610 (changes requested):
Even closer! Some things left. You shouldn't really need to change any existing test if we do this properly.

tests/conftest.py
herriejr: I modified this (locally) at the moment in mock_get_request to `data = payload if "v2" in url else {"meta": {"rc": "OK"}, "data": payload}`
Maybe passing in an extra property to mock_get_request, indicating the structure of the data, is a more future-proof way. But I'm missing some inspiration for the naming.

aiounifi/interfaces/traffic_rules.py
herriejr: Oops :blush:

aiounifi/interfaces/connectivity.py
herriejr: `{"errorCode":400,"message":"updateTrafficRule.arg2.matchingTarget: must not be null, updateTrafficRule.arg2.action: must not be null, updateTrafficRule.arg2.targetDevices: must not be empty"}`

This is the error I get for example when putting a wrong value to the endpoint for toggling a traffic rule. This is the 'raw' response, so that's everything we got to work with.

aiounifi/models/traffic_rule.py
herriejr: It depends on the request. If you do a TrafficRuleListRequest (so just 'getting' https://.../trafficrules) you get an array of the traffic rules present in the controller (or an empty array if none exist).
If you toggle a traffic rule, (so a put to https://.../trafficrules/{trafficrule_id}), it returns a JSON-object, so it is converted to a dict.
Another suggestion to get rid of this use of 'isinstance', is subclassing the TrafficRuleRequest with a new class TrafficRuleToggleRequest, that overrides the 'default' behaviour of the prepare_data-method of the TrafficRuleRequest.

aiounifi/models/api.py
herriejr: That's when you're in too deep... I did my reasoning the wrong way round. Since the structure of the trafficrules-response does not contain a meta- and data-key this wasn't working, so I overcomplicated things at the higher level.

aiounifi/interfaces/connectivity.py
herriejr: After doing the other modifications, the `and data["meta"] is not None` part is no longer needed.
But we still need to find a way to handle errors of the v2 API.

aiounifi/interfaces/api_handlers.py
herriejr: That's correct, but without this extra check I saw some errors in the tests. But I agree that it shouldn't be needed to change the tests if done properly. I'm also struggling with this a bit.
Since we now have a TypedApiResponse, maybe another method is needed, because we are not processing a 'raw' response anymore. I need to put some more time and thought into this.

aiounifi/models/api.py
Kane610: Well you got to a point where everything worked right. Bow we're just trying to fit everything together

aiounifi/models/traffic_rule.py
Kane610: I think doing a subclass per api request is better as its more clear what type of response is expected.

tests/conftest.py
Kane610: Sure, lets go with passing an extra property to mock_get_request where default behaviour goes towards current behaviour.

aiounifi/interfaces/api_handlers.py
Kane610: Well you would pass the data field into process raw which would still be a list[dict]] right?

aiounifi/interfaces/connectivity.py
Kane610: So suggestion is to keep the old logic as is and add a second if-statement that manages your data.

Kane610 (commented):
I think we're getting there! Try to take a step back and simplify what we've discussed and I think it's ready. I will need to rework a bit about where data response is parsed and error handling and stuff, it clearly doesn't work well with multiple data structures... But lets keep that out of this PR as much as possibly.

herriejr:
I have not much time in the coming days, but I will take a closer look at everything this weekend.

Kane610:
> I have not much time in the coming days, but I will take a closer look at everything this weekend.

Donit at your pace. I'm also away

Kane610:
What happened?

herriejr:
A bit of struggling with git and merging everything :scream:

Kane610:
> A bit of struggling with git and merging everything 😱
> I will reopen the PR shortly; no worries.

Omg! There are guides to revert these kind of mishaps. Haven't tried them though

herriejr:
Ok, back up and running. Just deciding on a way to handle the v2 api's.

Kane610:
I don't plan on doing any more changes on the library for now so you should be safe :)

herriejr:
> Ok, back up and running (I hope everything is fine). We just need to decide on a way to handle the v2 api errors.

For the moment, I have opted to put the v2 response in case of error in the format of the v1 response. There are also other possibilities to handle this. I have no preference, but it is just a first try.

Kane610:
> > Ok, back up and running (I hope everything is fine). We just need to decide on a way to handle the v2 api errors.
> 
> For the moment, I have opted to put the v2 response in case of error in the format of the v1 response. There are also other possibilities to handle this. I have no preference, but it is just a first try.

Excellent! I will look at it tomorrow evening.

aiounifi/models/traffic_rule.py
```diff
@@ -0,0 +1,187 @@
...
+@dataclass
+class TrafficRuleRequest(ApiRequest):
+    """Data class with required properties of a traffic rule API request."""
+
+    """We need a way to indicate if, for our model, the v2 API must be called.
+    Therefore an intermediate dataclass 'TrafficRuleRequest' is made,
+    for passing the correct path. This way, we do not need to alter any of the
+    other classes that not need to know about the version of the api used.
+    With the refactoring of the aiounifi-library, this is now possible.
+    """
```
Kane610, line 92:
```suggestion
    """Data class with required properties of a traffic rule API request.

    We need a way to indicate if, for our model, the v2 API must be called.
    Therefore an intermediate dataclass 'TrafficRuleRequest' is made,
    for passing the correct path. This way, we do not need to alter any of the
    other classes that not need to know about the version of the api used.
    With the refactoring of the aiounifi-library, this is now possible.
    """
```

aiounifi/interfaces/api_handlers.py
```diff
@@ -98,8 +98,15 @@ def __init__(self, controller: "Controller") -> None:
...
-        self.process_raw(raw.get("data", []))
+        data = await self.controller.request(self.api_request)
+        self.process_typed_api_response(data)
+
+    @final
+    def process_typed_api_response(self, response: TypedApiResponse) -> None:
+        """Process full typed API response"""
+        if "data" in response:
+            for response_item in response["data"]:
+                self.process_item(response_item)
```
Kane610, line 109:
```suggestion
        for response_item in response.get("data", []):
            self.process_item(response_item)
```

aiounifi/models/traffic_rule.py
```diff
@@ -0,0 +1,187 @@
...
+    ip_ranges: list[IPRange]
+    matching_target: str
+    network_ids: list[str]
+    regions: list[str]
+    schedule: Schedule
+    target_devices: list[TargetDevice]
+
+
+@dataclass
+class TrafficRuleRequest(ApiRequest):
```
Kane610, line 84:
```suggestion
class TrafficRuleRequest(ApiRequestV2):
```

How about we do an ApiRequestV2 in models.api that inherits from ApiRequest and add the general changes in this class to that one? Would make the class look just like the other ones

aiounifi/interfaces/api_handlers.py
```diff
@@ -98,8 +98,15 @@ def __init__(self, controller: "Controller") -> None:
     @final
     async def update(self) -> None:
         """Refresh data."""
-        raw = await self.controller.request(self.api_request)
-        self.process_raw(raw.get("data", []))
+        data = await self.controller.request(self.api_request)
+        self.process_typed_api_response(data)
```
Kane610, line 102: Is this change really necessary, I don't understand why this change is needed. From what I can tell you do exactly the same thing, process_raw only iterates over an iterator and then passes each item along to process_item

tests/conftest.py
```diff
@@ -99,16 +99,17 @@ def endpoint_fixture(
...
     """Use fixtures to mock all endpoints."""
 
     def mock_get_request(
-        path: str, unifi_path: str, payload: list[dict[str, Any]]
+        path: str, unifi_path: str, payload: list[dict[str, Any]], v2: bool = False
     ) -> None:
         """Register HTTP response mock."""
         url = unifi_path if is_unifi_os else path
-        data = {"meta": {"rc": "OK"}, "data": payload}
+        data = payload if v2 else {"meta": {"rc": "OK"}, "data": payload}
```
Kane610, line 112:
```suggestion
        data = payload if path.startswith("/v2") else {"meta": {"rc": "OK"}, "data": payload}
```
Not sure which is cleanest, your v2 option or just looking at the path

Kane610 (changes requested):
Good job! You're getting it into fantastic shape

Looks like you accidentally removed the test_outlet and test_port files

aiounifi/interfaces/traffic_rules.py
```diff
@@ -0,0 +1,37 @@
...
+        """Disable traffic rule defined in controller."""
+        return await self.toggle(traffic_rule, state=False)
+
+    async def toggle(self, traffic_rule: TrafficRule, state: bool) -> TypedApiResponse:
+        """Set traffic rule - defined in controller - to the desired state."""
+        traffic_rule_dict = deepcopy(traffic_rule.raw)
+        traffic_rule_response = await self.controller.request(
+            TrafficRuleEnableRequest.create(traffic_rule_dict, enable=state)
+        )
+        self.controller.traffic_rules.process_typed_api_response(traffic_rule_response)
```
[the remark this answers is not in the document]
herriejr: Done

aiounifi/models/traffic_rule.py
herriejr: Done

aiounifi/interfaces/api_handlers.py
herriejr: How could I miss that; that was so unpythonic of me :see_no_evil:

aiounifi/interfaces/api_handlers.py
herriejr: I think this has to do with me messing the merges up. This is indeed not needed.

aiounifi/interfaces/api_handlers.py
herriejr: I have been able to remove the 'process_typed_api_response', by calling process_raw with the correct argument.

aiounifi/interfaces/traffic_rules.py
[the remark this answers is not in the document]
herriejr: I have replaced this by `self.process_raw(traffic_rule_response.get("data", []))`, so there is no longer the need for the `process_typed_api_response`-function.

herriejr:
> Good job! You're getting it into fantastic shape
> 
> Looks like you accidentally removed the test_outlet and test_port files

I reverted those changes back, was a bit unfortunate of me.

tests/conftest.py
herriejr: For the v2 there are also 2 paths possible, depending if we are dealing with UnifiOS or not: `"/proxy/network/v2/api/site/{site}{self.path}"` and `"/v2/api/site/{site}{self.path}"`. 
I think your suggestion should then read `data = payload if "v2" in path else {"meta": {"rc": "OK"}, "data": payload}`
Both options work :wink: , but with the v2-boolean the responsibility for determining whether it is a v2 API or not is left to the user of the method. Therefore, this felt more natural for me.

aiounifi/models/traffic_rule.py
herriejr: Good suggestion, much cleaner and nicer. I added it in.

aiounifi/models/traffic_rule.py
```diff
@@ -0,0 +1,167 @@
...
+class TrafficRuleToggleRequest(ApiRequestV2):
+    """Data class with required properties of a traffic rule toggle API request."""
+
+    def prepare_data(self, raw: bytes) -> TypedApiResponse:
+        """Put data, received from the unifi controller, into a TypedApiResponse."""
+        json_data = orjson.loads(raw)
+        return_data: TypedApiResponse = {}
+        return_data["meta"] = self.handle_error(json_data)
+        return_data["data"] = [json_data]
+        return return_data
```
Kane610, line 106:
```suggestion
    def prepare_data(self, raw: bytes) -> TypedApiResponse:
        """Put data, received from the unifi controller, into a TypedApiResponse."""
        json_data = orjson.loads(raw)
        return_data: TypedApiResponse = {}
        return_data["meta"] = self.handle_error(json_data)
        return_data["data"] = json_data
        return return_data


@dataclass
class TrafficRuleToggleRequest(ApiRequestV2):
    """Data class with required properties of a traffic rule toggle API request."""

    def prepare_data(self, raw: bytes) -> TypedApiResponse:
        """Put data, received from the unifi controller, into a TypedApiResponse."""
        json_data = orjson.loads(raw)
        return_data: TypedApiResponse = {}
        return_data["meta"] = self.handle_error(json_data)
        return_data["data"] = [json_data]
        return return_data
```
I think this very well also can be part of the V2 base class, you only need to consider json_data being a list or not right? Does it being a list only depend on if it is a get or post?

tests/conftest.py
Kane610: Well both path and unifi_path would be a v2 path right? That's why I'm only suggesting looking at the path string which starts with V2

aiounifi/models/api.py
```diff
@@ -27,6 +29,36 @@ def full_path(self, site: str, is_unifi_os: bool) -> str:
             return f"/proxy/network/api/s/{site}{self.path}"
         return f"/api/s/{site}{self.path}"
 
+    def prepare_data(self, raw: bytes) -> TypedApiResponse:
```
Kane610, line 32:
```suggestion
    def decode(self, raw: bytes) -> TypedApiResponse:
```

Lets rename this to decode.

I wonder if we should introduce a TypedApiResponseV2 that handles converting the response to a common format. Lets not do this now...

tests/test_ports.py
```diff
@@ -7,6 +7,7 @@
 import pytest
 
 from aiounifi.interfaces.api_handlers import ItemEvent
+from aiounifi.models.api import TypedApiResponse
```
Kane610, line 10:
```suggestion
```

tests/test_outlets.py
```diff
@@ -8,6 +8,7 @@
 import pytest
 
 from aiounifi.interfaces.api_handlers import ItemEvent
+from aiounifi.models.api import TypedApiResponse
```
Kane610, line 11:
```suggestion
```

Kane610 (changes requested):
I like your improvements it fits well in with the overall structure.

Only more would be to handle response differences in the response rather than request class, but that's more work and let's keep out of it.

aiounifi/models/api.py
```diff
@@ -27,6 +29,36 @@ def full_path(self, site: str, is_unifi_os: bool) -> str:
...
+    """
+
+    def full_path(self, site: str, is_unifi_os: bool) -> str:
+        """Create url to work with a specific controller."""
+        if is_unifi_os:
+            return f"/proxy/network/v2/api/site/{site}{self.path}"
+        return f"/v2/api/site/{site}{self.path}"
+
+    def handle_error(self, json: dict[str, Any]) -> dict[str, Any]:
+        if "errorCode" in json:
```
Kane610, line 56: Missing docstring

aiounifi/models/api.py
```diff
@@ -27,6 +29,36 @@ def full_path(self, site: str, is_unifi_os: bool) -> str:
...
+    def prepare_data(self, raw: bytes) -> TypedApiResponse:
+        """Put data, received from the unifi controller, into a TypedApiResponse."""
+        return_data: TypedApiResponse = orjson.loads(raw)
+        return return_data
+
+
+@dataclass
+class ApiRequestV2(ApiRequest):
+    """Data class with required properties of a V2 API request.
+    We need a way to indicate if, for our model, the v2 API must be called.
```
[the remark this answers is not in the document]
herriejr: Done, changed it back :wink:

aiounifi/models/api.py
herriejr: Fixed.

aiounifi/models/api.py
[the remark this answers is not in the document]
Kane610: I think you actually had two strings before.

aiounifi/models/api.py
herriejr: Ok, decode is a much better name; propagated this in the rest of the code of course.

Maybe in the future a conversion to a common format would be interesting. I agree that introducing this now, probably would take us too far.

aiounifi/models/traffic_rule.py
herriejr: Yes, I only need to consider json_data being a list or not.
A GET results in a list.
A PUT results in a dict.

I moved this 'prepare_data' (now 'decode') to the V2 base class; this also enabled me to remove the other TrafficRule...Request classes. I will commit this in a few moments.

Kane610:
I glanced from the phone and it looks really good. Will review tomorrow

Kane610 (approved):
Im very happy with how this turned out. You did a great job!

Kane610:
I saw one thing that warrants a fix, test coverage

`aiounifi/interfaces/traffic_rules.py               18      6    67%   24, 28, 32-37`

tests/test_traffic_rules.py
```diff
@@ -0,0 +1,67 @@
...
+    mock_aioresponse.put(
+        "https://host:8443/proxy/network/v2/api/site/default/trafficrules/6452cd9b859d5b11aa002ea1",
+        payload={},
+    )
+
+    traffic_rule = TRAFFIC_RULES[0]
+
+    await unifi_controller.request(
+        TrafficRuleEnableRequest.create(traffic_rule, enable)
+    )
```
Kane610, line 28: Change this to use the three control methods of Traffic and you're home :)

herriejr: I left this in, but I added additional tests to use the three control methods.

Kane610: Perfect! Will review tomorrow

Kane610 (approved):
Looks good! 🎉 

Anything more you want to do here or are you ready to get this merged?

herriejr:
I am ready to get this merged. Thank you for your time and effort for guiding me.

ViViDboarder:
I stumbled on this PR as I was looking to do something like it myself. Not only was I glad to see it was done (yesterday, no less!), I found this PR to be overall super uplifting.

@Kane610 your feedback and attitude was terrific and inspirational.

@herriejr your persistence and open mindedness to learn was impressive. Really great job!

Reading this made me feel good about open source, so I figured I’d share so you two can feel good about being a part of it. Kudos and thank you!

Kane610:
Thank you very much @ViViDboarder <3

Kane610:
@herriejr FYI I've published aiounifi v64 as I needed to propagate a fix to home assistant.

Kane610:
@herriejr I realised there is no tests with errors, Im reworking error handling a bit and would like to use that one better. Could you provide me with a couple of error examples for traffic rules (or rather v2 APIs)

herriejr:
I will be away for work the upcoming week. I will then look into it.

Kane610:
> I will be away for work the upcoming week. I will then look into it.

Thank you. That will be fine

PR #430, opened 2023-08-10
Adds missing outlet_ac fields to Device model

Cisien:
Currently `outlet_ac_power_budget` and `outlet_ac_power_consumption` are only exposed via the raw data. This change exposes these fields as properties on the Device model.

### DIFF
--- a/aiounifi/models/device.py
+++ b/aiounifi/models/device.py
@@ -754,6 +754,16 @@ class Device(ApiItem):
         return self.raw.get("outlet_overrides", [])
 
     @property
+    def outlet_ac_power_budget(self) -> str | None:
+        """The amount of power available to outlets."""
+        return self.raw.get("outlet_ac_power_budget")
+
+    @property
+    def outlet_ac_power_consumption(self) -> str | None:
+        """The amount of power consumed by all outlets."""
+        return self.raw.get("outlet_ac_power_consumption")
+
+    @property
     def outlet_table(self) -> list[TypedDeviceOutletTable]:
         """List of outlets."""
         return self.raw.get("outlet_table", [])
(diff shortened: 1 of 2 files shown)


aiounifi/models/device.py
```diff
@@ -753,6 +753,16 @@ def outlet_overrides(self) -> list[TypedDeviceOutletOverrides]:
         """Overridden outlet configuration."""
         return self.raw.get("outlet_overrides", [])
 
+    @property
+    def outlet_ac_power_budget(self) -> str | None:
+        """The amount of power available to outlets."""
+        return self.raw.get("outlet_ac_power_budget", None)
```
Kane610, line 759:
```suggestion
        return self.raw.get("outlet_ac_power_budget")
```
None is default return value in get

aiounifi/models/device.py
```diff
@@ -753,6 +753,16 @@ def outlet_overrides(self) -> list[TypedDeviceOutletOverrides]:
...
+
+    @property
+    def outlet_ac_power_consumption(self) -> str | None:
+        """The amount of power consumed by all outlets."""
+        return self.raw.get("outlet_ac_power_consumption", None)
```
Kane610, line 764:
```suggestion
        return self.raw.get("outlet_ac_power_consumption")
```

Kane610 (changes requested):

tests/test_devices.py
```diff
@@ -440,6 +440,8 @@ async def test_device_pdu_pro(mock_aioresponse, unifi_controller, unifi_called_w
     assert pdupro.model == "USPPDUP"
     assert pdupro.name == "Main Server Cabinet PDU"
     assert pdupro.next_interval == 56
+    assert pdupro.outlet_ac_power_budget =="1875.000"
```
Kane610, line 443:
```suggestion
    assert pdupro.outlet_ac_power_budget == "1875.000"
```

Kane610 (changes requested):

Kane610 (approved):
Awesome

Kane610:
I will do a release tonight for you to use with your integration PR.

Issue #522, opened 2023-11-11
Adding Traffic routes


shaunpearsondev:
Hi @Kane610 

I am looking at adding functionality for Traffic routes to this library and would like to get some of your time to get your thoughs on this before I go ahead and code it all up, the functionality seems to be behind the v2 api (like traffic rules) 

Here is a very early draft:
https://github.com/bigwhitesolutions/aiounifi/commit/1ea853f83856188651736f30f6ae29844a972044

It has very simular models to traffic rules so wondering if these should be shared?

Thank you!

Shaun

Kane610:
Hello and welcome!

Put up a PR and we can cooperate on the details. Skimmed through your linked code and from that quick glance it looked OK.

Let's keep all endpoints standalone as much as possible in case it changes later on. Keeping it explicit helps with future refactoring

Kane610:
This is completed

PR #528, opened 2023-11-18
Add traffic routes

shaunpearsondev:
As mentioned in #522 This adds the ability to list and toggle traffic routes

Very heavily follows the work on traffic rules 

Tested against 

UDM SE v3.1.16
Network 7.5.187

```python
result = await controller.request(TrafficRouteListRequest.create())
LOGGER.info(result)

 ```
```python
route = controller.traffic_routes["_id"]
result = await controller.request(
        TrafficRouteEnableRequest.create(deepcopy(route.raw), False)
    )
LOGGER.info(result)
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -70,3 +70,6 @@ venv.bak/
 .ruff_cache
 
 .vscode/settings.json
+
+#Jetbrains
+.idea
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,6 @@
+{
+    "recommendations": [
+        "charliermarsh.ruff",
+        "ms-python.python"
+    ]
+}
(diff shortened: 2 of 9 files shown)


Kane610:
Please fix so tests pass. Preferably you utilise the setup.sh script to get pee commit hools going as well.

Looked through the code and it looks good (indeed very much similar to traffic rules). Gonna look through it again tomorrow

shaunpearsondev:
Sorry windows user here 😳

I ran the following adapted script and all tests and pre commit hooks seem to be passing except the `test_websocket` one I can't see where I would of broken this so hoping it is a windows quirk...

```powershell
python3 -m venv venv

./venv/Scripts/Activate.ps1

python3 -m pip install --upgrade pip

python3 -m pip install .[requirements,requirements_test,requirements_dev]

pre-commit install
```

Kane610:
Thank you very much!

Would you mind doing a follow up PR that introduces a couple of test cases that has some errors from the API, would be beneficial for this https://github.com/Kane610/aiounifi/pull/520

PR #529, opened 2023-11-19
Add traffic routes

ViViDboarder:
Adds support for reading, enabling, disabling, as well as generally updating Traffic Routes.

I based this off the PR #409

One thing I wasn't sure about was what fields I should make properties vs leaving as accessible via `.raw`. I made a few of the ones that I plan to use, but I'm happy to just make them all accessible that way.

PR #537, opened 2023-11-22
Add some more typing to Traffic Routes and save

ViViDboarder:
Adds some more types and makes the saving function more explicit.

Rebased from #529

### DIFF
--- a/aiounifi/models/traffic_route.py
+++ b/aiounifi/models/traffic_route.py
@@ -59,8 +59,8 @@ class TypedTrafficRoute(TypedDict):
     """Traffic route type definition."""
 
     _id: str
-    description: str
     action: str
+    description: str
     domains: list[Domain]
     enabled: bool
     ip_addresses: list[IPAddress]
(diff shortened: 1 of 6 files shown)


Kane610 (changes requested):

Kane610 (changes requested):
The points I marked duplicate, make sure they're sorted alphabetically as well. I think the ones I marked was the duplicate in the wrong order.

Kane610:
Awesome! Only two things left

aiounifi/models/traffic_route.py
```diff
@@ -88,21 +114,59 @@ def id(self) -> str:
...
     @property
     def description(self) -> str:
         """Description given by user to traffic route."""
         return self.raw["description"]
 
+    @property
+    def domains(self) -> list[Domain]:
+        """What IP addresses are matched against by this traffic route.
+
+        Note: This requires clients to use the UniFi network devices as their DNS server.
```
Kane610, line 131:
```suggestion
        Note: This requires clients to use the UniFi network devices as their DNS server.
```
This line is a bit weird to me, with devices you mean the UniFi gateway/router right?

Kane610 (changes requested):
Final thing Please increase test coverage of your additions to 100%

aiounifi/models/traffic_route.py
ViViDboarder: I'll reword this to be the exact message from the documentation: `Note: Domain matching requires the client devices to use the UniFi gateway as the DNS server.`

ViViDboarder:
Should be up to 100% now.

tests/test_traffic_routes.py
```diff
@@ -104,8 +110,14 @@ async def test_traffic_routes(unifi_controller, _mock_endpoints, unifi_called_wi
...
     assert traffic_route.description == "TEST 1"
+    assert traffic_route.domains == []
     assert traffic_route.enabled is False
-    assert traffic_route.matching_target == "INTERNET"
+    assert traffic_route.ip_addresses == []
+    assert traffic_route.ip_ranges == []
+    assert traffic_route.matching_target == MatchingTarget.INTERNET
+    assert traffic_route.network_id == "6547e87f6e34654e72293637"
+    assert traffic_route.next_hop == ""
+    assert traffic_route.regions == []
```
Kane610, line 120: Since these are very specific it would be good if the data sets where not empty

Kane610 (changes requested):
One final thing

ViViDboarder:
So, this diff ends up changing more. I realized the other test was modifying the fixtures values which munged with other tests. So I made a change to make sure it's a deep copy. That broke other tests so I added some more asserts and made them more robust.

Kane610 (approved):
Awesome

PR #548, opened 2023-12-08
Add IntEnum and dictionary for device states

JosephBlock:
Add IntEnum and dictionary for device states to support additional diagnostics in HA
[#105138](https://github.com/home-assistant/core/pull/105138)

### DIFF
--- a/aiounifi/models/device.py
+++ b/aiounifi/models/device.py
@@ -4,6 +4,7 @@ Access points, Gateways, Switches.
 """
 from copy import deepcopy
 from dataclasses import dataclass
+from enum import IntEnum
 import logging
 from typing import Any, NotRequired, Self, TypedDict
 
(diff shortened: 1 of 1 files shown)


Kane610 (changes requested):
Also update the state property return value to be the DeviceState

aiounifi/models/device.py
```diff
@@ -843,7 +835,7 @@ def port_table(self) -> list[TypedDevicePortTable]:
         return self.raw.get("port_table", [])
 
     @property
-    def state(self) -> int:
+    def state(self) -> DeviceState:
         """State of device."""
         return self.raw["state"]
```
Kane610, line 840:
```suggestion
        return DeviceState(self.raw["state"])
```

Kane610 (approved):

Kane610:
Awesome! I will look to do a release of aiounifi shortly

PR #562, opened 2024-01-03
feat: add typing for Speedtest Status

owenvoke:
I'm not fully sure whether this is the right way to add this, but it'd be really nice to have these values typed so that they could be added to Home Assistant (rather than having to use `speedtest.net` as well). 🙂

### DIFF
--- a/aiounifi/models/device.py
+++ b/aiounifi/models/device.py
@@ -852,8 +852,7 @@ class Device(ApiItem):
     @property
     def speedtest_status(self) -> TypedDeviceSpeedtestStatus | None:
         """Speedtest status."""
-        value = self.raw.get("speedtest-status")
-        if value is None:
+        if (value := self.raw.get("speedtest-status")) is None:
             return None
         return cast(TypedDeviceSpeedtestStatus, value)
 
--- a/aiounifi/models/device.py
+++ b/aiounifi/models/device.py
@@ -6,7 +6,7 @@ from copy import deepcopy
 from dataclasses import dataclass
 from enum import IntEnum
 import logging
-from typing import Any, NotRequired, Self, TypedDict
+from typing import Any, NotRequired, Self, TypedDict, cast
 
 from .api import ApiItem, ApiRequest
 
@@ -338,6 +338,20 @@ class TypedDeviceWlanOverrides(TypedDict):
     wlan_id: str
 
 
+class TypedDeviceSpeedtestStatus(TypedDict):
+    """Device speedtest status type definition."""
+
+    latency: int
+    rundate: int
+    runtime: int
+    status_download: int
+    status_ping: int
+    status_summary: int
+    status_upload: int
+    xput_download: float
+    xput_upload: float
+
+
 class TypedDevice(TypedDict):
     """Device type definition."""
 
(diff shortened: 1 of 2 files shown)


aiounifi/models/device.py
```diff
@@ -834,6 +849,14 @@ def port_table(self) -> list[TypedDevicePortTable]:
         """List of ports and data."""
         return self.raw.get("port_table", [])
 
+    @property
+    def speedtest_status(self) -> TypedDeviceSpeedtestStatus | None:
+        """Speedtest status."""
+        value = self.raw.get("speedtest-status")
+        if value is None:
+            return None
+        return cast(TypedDeviceSpeedtestStatus, value)
```
owenvoke, line 858: This is the bit I'm not sure about. If I leave it as just `return self.raw.get("speedtest-status")`, it complains about the return type not matching. 🤷🏻

Kane610: That is because the typeddict can't handle the '-' and instead its added as a '_'.

Check system_stats in the device which has the same issue

owenvoke: Is it ok as-is then? 🙂

owenvoke: Convert to a tuple instead of a typed dict? 🤔 Happy to make that change.

```suggestion
        data = self.raw.get("speedtest-status")
        if data is None:
            return None
        return (
            data.get("latency", 0),
            data.get("rundate", 0),
            data.get("runtime", 0),
            data.get("status_download", 0),
            data.get("status_ping", 0),
            data.get("status_summary", 0),
            data.get("status_upload", 0),
            data.get("xput_download", 0),
            data.get("xput_upload", 0),
        )
```

Or do you mean removing the `None` check? 🤔 `speedtest-status` won't be included for most devices, so that might not be possible.

Kane610:
```suggestion
        if (value := self.raw.get("speedtest-status")) is None:
            return None
        return cast(TypedDeviceSpeedtestStatus, value)
```
Sorry, that was a bad example as it didn't do what I thought, small suggestion on improving your original PR

owenvoke: Ah, didn't know you could do that. Looks much cleaner!

aiounifi/models/device.py
```diff
@@ -338,6 +338,20 @@ class TypedDeviceWlanOverrides(TypedDict):
...
+
+    latency: int
+    rundate: int
+    runtime: int
+    status_download: int
+    status_ping: int
+    status_summary: int
+    status_upload: int
+    xput_download: float
+    xput_upload: float
```
Kane610, line 352: I just have one question, is this still the same on the latest versions of the network app?

aiounifi/models/device.py
Kane610: There's a lot of nice additions to the python language over the last few releases.

aiounifi/models/device.py
owenvoke: Yes, for me it is (UDM SE, Network  v8.0.26).

Kane610: Awesome!

Kane610:
I will put up a new release in the next few days

Kane610:
aiounifi v69 has been released and is now usable in the integration

PR #593, opened 2024-02-21
feat: adds support for getting led_override statuses for devices

layertwo:
### Overview 

I'm working on adding LED control for Unifi devices in HA, but needed to expose the LED control in the underlying library.

Some interesting notes:
- Not all devices support RGB LED, and therefore need to ensure the outgoing API call does not include the fields for that (although I'm not sure Unifi actually cares)
- There's a hardware capabilities field (`device.hw_caps`) that can be used to determine if the device supports brightness / color. Reference from [here](https://github.com/misenhower/homebridge-unifi-led-control/blob/master/src/UniFiDevice.js#L42-L44). I'm not sure how accurate that is, but out of the 5 devices I have home my network, only 1 supports RGB, and my testing proves this is accurate.

### Testing

- New unit tests support both the model change and new `DeviceSetLedStatus` request method.
- Manual testing against my Unifi devices


Kane610:
Please setup (setup.sh) the environment so you get ruff, ruff-check and mypy in order when publishing PRs

aiounifi/models/device.py
```diff
@@ -917,6 +953,26 @@ def wlan_overrides(self) -> list[TypedDeviceWlanOverrides]:
         """Wlan configuration override."""
         return self.raw.get("wlan_overrides", [])
 
+    @property
+    def supports_led_ring(self) -> bool:
+        """Check if the hardware supports an LED ring based on the second bit of `hw_caps`."""
+        return bool(self.raw.get("hw_caps", 0) & (1 << 1))
```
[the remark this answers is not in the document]
layertwo: So made `hw_caps` its own property. 

What are you looking for with the constants? I'm not exactly sure how UBNT encodes their hardware capabilities as an int. 

Found the reference in some code for HomeBridge that seemed to be usable. 

I can add some docstrings with some sample values if that would be useful.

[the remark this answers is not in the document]
Kane610: I would like something like https://github.com/home-assistant/core/blob/e88dfd4f6870644243c8737ed29ec0ff4df7fada/homeassistant/components/light/__init__.py#L48

Which you can then use with a bitwise or https://wiki.python.org/moin/BitwiseOperators

Similar to this https://github.com/home-assistant/core/blob/e88dfd4f6870644243c8737ed29ec0ff4df7fada/homeassistant/components/deconz/light.py#L196

The end result would be something like bool(self.hw_caps & HwCapability.LED_RING)

The end result will be the same but much more understandable

aiounifi/models/device.py
```diff
@@ -722,6 +723,43 @@ def create(
...
+            else:
+                raise AttributeError("Brightness must be within the range [0, 100].")
+
+            # Validate color parameter
+            if color is not None and re.match(r"^#(?:[0-9a-fA-F]{3}){1,2}$", color):
+                data["led_override_color"] = color
+            else:
+                raise AttributeError(
+                    "Color must be a valid hex color code (e.g., '#00FF00')."
+                )
```
Kane610, line 754: Can you send brightness or led_override_color without the other?

If so then I think this would suffice?
```suggestion
            if brightness is not None and (0 <= brightness <= 100):
                data["led_override_color_brightness"] = brightness

            # Validate color parameter
            if color is not None and re.match(r"^#(?:[0-9a-fA-F]{3}){1,2}$", color):
                data["led_override_color"] = color
```

Kane610, line 754: Is it required it send both brightness and led_override_color?

```suggestion
        if device.supports_led_ring and brightness is not None and color is not None:
            # Validate brightness parameter
            if 0 <= brightness <= 100:
                data["led_override_color_brightness"] = brightness
            else:
                raise AttributeError("Brightness must be within the range [0, 100].")

            # Validate color parameter
            if re.match(r"^#(?:[0-9a-fA-F]{3}){1,2}$", color):
                data["led_override_color"] = color
            else:
                raise AttributeError(
                    "Color must be a valid hex color code (e.g., '#00FF00')."
                )
```

Kane610 (changes requested):

aiounifi/models/device.py
layertwo: I don't believe it's required to send both together. Can double check against my environment, but I don't believe you need to send both.

layertwo: Answered below

aiounifi/models/device.py
```diff
@@ -722,6 +723,49 @@ def create(
...
+    ) -> Self:
+        """Set LED status of device."""
+
+        data = {"led_override": status}
+        if device.supports_led_ring:
+            # Validate brightness parameter
+            if brightness is not None and (0 <= brightness <= 100):
+                data["led_override_color_brightness"] = brightness
+            else:
+                raise AttributeError("Brightness must be within the range [0, 100].")
```
Kane610, line 746: If you don't need to send both then you need to rework this
```suggestion
            if brightness is not None: 
                if (0 <= brightness <= 100):
                    data["led_override_color_brightness"] = brightness
                else:
                    raise AttributeError("Brightness must be within the range [0, 100].")
```

aiounifi/models/device.py
```diff
@@ -722,6 +723,49 @@ def create(
...
+
+            # Validate color parameter
+            if color is not None and re.match(r"^#(?:[0-9a-fA-F]{3}){1,2}$", color):
```
Kane610, line 749: Same as brightness need to be reworked if you support either one option to be sent.

aiounifi/models/device.py
```diff
@@ -917,6 +961,35 @@ def wlan_overrides(self) -> list[TypedDeviceWlanOverrides]:
...
+        """LED override color - if supports LED ring."""
+        if self.supports_led_ring:
+            return self.raw.get("led_override_color")
+
+    @property
+    def led_override_color_brightness(self) -> int | None:
+        """LED override color brightness - if supports LED ring."""
+        if self.supports_led_ring:
+            value = self.raw.get("led_override_color_brightness")
+            assert isinstance(value, int)
```
Kane610, line 990:
```suggestion
            value = self.raw.get("led_override_color_brightness")
```
I saw a typo (`led_override_color_brightnes: int`) in `class TypedDevice(TypedDict):` if you add another s in brightness you shouldn't need to use the assert

tests/test_devices.py
```diff
@@ -999,3 +1004,99 @@ async def test_device_websocket(
 def test_enum_unknowns() -> None:
     """Validate enum unknown values."""
     assert DeviceState(999) == DeviceState.UNKNOWN
+
+
+@pytest.mark.parametrize(
+    ("device_payload", "data", "command"),
+    [
```
Kane610, line 1011: Also add test coverage for the cases where you only use either brightness or color

aiounifi/models/device.py
```diff
@@ -722,6 +723,49 @@ def create(
...
+    def create(
+        cls,
+        device: "Device",
+        status: str = "on",
+        brightness: int | None = None,
+        color: str | None = None,
+    ) -> Self:
+        """Set LED status of device."""
+
+        data = {"led_override": status}
```
Kane610, line 740:
```suggestion
        data: dict[str, int | str] = {"led_override": status}
```
You also put integers into data so you need to specify the possible types

aiounifi/models/device.py
```diff
@@ -917,6 +961,35 @@ def wlan_overrides(self) -> list[TypedDeviceWlanOverrides]:
...
+        if self.supports_led_ring:
+            return self.raw.get("led_override_color")
+
+    @property
+    def led_override_color_brightness(self) -> int | None:
+        """LED override color brightness - if supports LED ring."""
+        if self.supports_led_ring:
+            value = self.raw.get("led_override_color_brightness")
+            assert isinstance(value, int)
+            return value
```
Kane610, line 991: I saw a typo (led_override_color_brightnes: int) in class TypedDevice(TypedDict): if you add another s in brightness you shouldn't need to use the assert


```suggestion
        return self.raw.get("led_override_color_brightness")
```
The self.supports_led_ring is a convenience method for higher functionality, you can just do a get on the key itself

aiounifi/models/device.py
```diff
@@ -917,6 +961,35 @@ def wlan_overrides(self) -> list[TypedDeviceWlanOverrides]:
...
+    @property
+    def led_override(self) -> str | None:
+        """LED override."""
+        return self.raw.get("led_override")
+
+    @property
+    def led_override_color(self) -> str | None:
+        """LED override color - if supports LED ring."""
+        if self.supports_led_ring:
+            return self.raw.get("led_override_color")
```
Kane610, line 983:
```suggestion
        return self.raw.get("led_override_color")
```
The self.supports_led_ring is a convenience method for higher functionality, you can just do a get on the key itself

Kane610 (changes requested):
Nice improvements, nearly there!

layertwo:
Alright, addressed comments, should be good to go.

Kane610 (approved):
Awesome! Good job

dziedzic99:
Hey @layertwo
How's that HA integration going? I was thinking of picking this up. Any updates?

layertwo:
@dziedzic99 I think I have some WIP code on my desktop in some state, but it's been over 6 months since I last looked at it.

I can commit it you want to use it as a starting point. I think when I left it, it just needed unit tests.

dziedzic99:
@layertwo 
Yes please, I will happily test it in some spare time.

PR #618, opened 2024-03-15
Only use HTTPS if ssl_context exists

mindphluxnet:
Despite being able to set a SSL context the URI used to connect to the controller is hardcoded to use HTTPS. Not every controller runs on HTTPS so connecting will always fail with the error

> 2024-03-15 21:17:48.964 ERROR (MainThread) [homeassistant.components.unifi] Error connecting to the UniFi Network at uinet: Error requesting data from https://host:8443: Cannot connect to host host:8443 ssl:False [Name has no usable address]

This feels like an oversight so my PR fixes this.

### DIFF
--- a/aiounifi/models/configuration.py
+++ b/aiounifi/models/configuration.py
@@ -23,4 +23,8 @@ class Configuration:
     @property
     def url(self) -> str:
         """Represent console path."""
-        return f"https://{self.host}:{self.port}"
+        return (
+            f"https://{self.host}:{self.port}"
+            if self.ssl_context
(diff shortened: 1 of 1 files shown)


Kane610:
Thanks for contributing but I don't see the need for this.

People have been running it in all types of setups the last ~6 years and never have anyone complained. Why I ever passed in the context was if I would ever handle proper certificates which no-one also never complained on. My current plan is to remove the context and clean that part up.

mindphluxnet:
Alright.

PR #640, opened 2024-04-17
Handle login non JSON response

mkmer:
This handles the non JSON response and causes retries until the host is "alive"
There may be a better way to handle things.  I selected .debug for the log as it "spams" the log file with errors during a "normal" startup

### DIFF
--- a/aiounifi/interfaces/connectivity.py
+++ b/aiounifi/interfaces/connectivity.py
@@ -74,7 +74,7 @@ class Connectivity:
                 LOGGER.error("Login failed '%s'", data)
                 raise ERRORS.get(data["meta"]["msg"], AiounifiException)
         else:
-            LOGGER.error("Login Failed not JSON: '%s'", bytes_data)
+            LOGGER.debug("Login Failed not JSON: '%s'", bytes_data)
             raise RequestError("Login Failed: Host starting up")
   
         if (
--- a/aiounifi/interfaces/connectivity.py
+++ b/aiounifi/interfaces/connectivity.py
@@ -73,15 +73,17 @@ class Connectivity:
             if "meta" in data and data["meta"]["rc"] == "error":
                 LOGGER.error("Login failed '%s'", data)
                 raise ERRORS.get(data["meta"]["msg"], AiounifiException)
+
+        elif response.content_type == "text/html":
+            LOGGER.debug("Login Failed not JSON: '%s'", bytes_data)
+            raise RequestError("Login Failed: Host starting up")
+
         if (
             response.status == HTTPStatus.OK
             and (csrf_token := response.headers.get("x-csrf-token")) is not None
         ):
             self.headers["x-csrf-token"] = csrf_token
 
-        elif response.content_type == "text/html":
-            LOGGER.debug("Login Failed not JSON: '%s'", bytes_data)
-            raise RequestError("Login Failed: Host starting up")
 
 
         self.can_retry_login = True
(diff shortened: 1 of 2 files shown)


mkmer:
It looks like the tests are expecting HTML for this part:
https://github.com/Kane610/aiounifi/blob/5452ff3eee2279d1ed988abb396e8493802da120/tests/test_controller.py#L41

Not sure if it's a "bad" test, real world testing with my controller makes it through the startup and eventually connects.

mkmer:
Sorry about all of the "noise" - this should be the "final answer" - original tests still pass, new test covers the new code AND it actually fixes the startup problem :)

Kane610:
> Sorry about all of the "noise" - this should be the "final answer" - original tests still pass, new test covers the new code AND it actually fixes the startup problem :)

Awesome! Don't worry about it. I looked through it but will review tonight

aiounifi/interfaces/connectivity.py
```diff
@@ -74,6 +74,10 @@ async def login(self) -> None:
                 LOGGER.error("Login failed '%s'", data)
                 raise ERRORS.get(data["meta"]["msg"], AiounifiException)
 
+        elif response.content_type == "text/html":
```
Kane610, line 77: I guess we could just do an else here. But let's keep it like this for now

Kane610 (approved):

Kane610:
Thanks

Kane610:
Have you validated that this works in reality?

aiounifi/interfaces/connectivity.py
mkmer: I started with an else here, and the tests failed - I don't recall which one, but we do need to look for text/html - I think there is a valid text/json header in the tests.

mkmer:
Yes, I tested with my configuration.  As long as the host is returning "starting" with text/html the old code failed, this code triggers HA to retry the integration until it returns the "correct" login info.

aiounifi/interfaces/connectivity.py
Kane610: Some old tests are not correct so there are room for improvement, maybe Ill add a warning in an else and see if someone reports it :)

Kane610:
Great! Will put up a new release tonight

aiounifi/interfaces/connectivity.py
mkmer: I think it needs a bit more digging.  There are 2 if statements - else on the first one: 
https://github.com/Kane610/aiounifi/blob/8afb578803d4462fbb1ba0a635437283ccfcc70b/aiounifi/interfaces/connectivity.py#L71 
caused use to miss the second test: 
https://github.com/Kane610/aiounifi/blob/8afb578803d4462fbb1ba0a635437283ccfcc70b/aiounifi/interfaces/connectivity.py#L81
Else on the second test was not good because the response is `HTTPStatus.OK`, just the wrong content_type.
If we move the second test around - or know that the `x-crf-token` is never found in a text/html or text/json.

Reviewing one "source" https://stackoverflow.com/questions/23714383/what-are-all-the-possible-values-for-http-content-type-header it appears text/json is not even a valid content type?  If that's so, then we should only look for the `x-crf-token` in `application/json` and use an else.

Sometimes it amazes me how much thought can go into a couple of `if` statements!

mkmer: I did a quick PR https://github.com/Kane610/aiounifi/pull/641 based on what I think you were thinking :)

PR #641, opened 2024-04-18
Rework test content type / refactor if tests in connection/login

mkmer:
After reviewing the possible content types, I believe the `text/json` should be `application/json`.  After making the change, all original code passes tests.  This allows for a bit of refactoring in the login `if` chain that may be more like what you were thinking.

This also "works" - no real need to change it, but I think it is a bit easier to read?

### DIFF
--- a/aiounifi/interfaces/connectivity.py
+++ b/aiounifi/interfaces/connectivity.py
@@ -74,15 +74,12 @@ class Connectivity:
                 LOGGER.error("Login failed '%s'", data)
                 raise ERRORS.get(data["meta"]["msg"], AiounifiException)
 
-        elif response.content_type == "text/html":
+            if (csrf_token := response.headers.get("x-csrf-token")) is not None:
+                self.headers["x-csrf-token"] = csrf_token
+
+        else:
             LOGGER.debug("Login Failed not JSON: '%s'", bytes_data)
             raise RequestError("Login Failed: Host starting up")
-
-        if (
-            response.status == HTTPStatus.OK
-            and (csrf_token := response.headers.get("x-csrf-token")) is not None
-        ):
-            self.headers["x-csrf-token"] = csrf_token
 
         self.can_retry_login = True
         LOGGER.debug("Logged in to UniFi %s", url)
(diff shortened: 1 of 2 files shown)


aiounifi/interfaces/connectivity.py
```diff
@@ -74,16 +74,13 @@ async def login(self) -> None:
                 LOGGER.error("Login failed '%s'", data)
                 raise ERRORS.get(data["meta"]["msg"], AiounifiException)
 
-        elif response.content_type == "text/html":
+            if (csrf_token := response.headers.get("x-csrf-token")) is not None:
+                self.headers["x-csrf-token"] = csrf_token
+
+        else:
             LOGGER.debug("Login Failed not JSON: '%s'", bytes_data)
             raise RequestError("Login Failed: Host starting up")
```
Kane610, line 83: This can still be done after the if/else
```suggestion
        else:
            LOGGER.debug("Login Failed not JSON: '%s'", bytes_data)
            raise RequestError("Login Failed: Host starting up")

        if (csrf_token := response.headers.get("x-csrf-token")) is not None:
            self.headers["x-csrf-token"] = csrf_token
```

mkmer: I suppose it could - it was broken with the tests having `text/json` - If we move it out do we need to put the `response.status == HTTPStatus.OK` test back in line 81?  I think so....

Kane610: I'm not sure. I would have to check the commit history

Kane610: The code looks quite a bit different today from when it was added, so I think its okay without the check

https://github.com/Kane610/aiounifi/pull/140/files

Kane610: Im gonna merge this so I can get out a fix for 2024.4.4.

mkmer: I think it's ok either way - at least it knows when a login failed :)
Thanks for all your work on this "fun" stuff.

PR #649, opened 2024-04-23
Clean up login logic

Kane610:
Follow up to https://github.com/Kane610/aiounifi/pull/640 and https://github.com/Kane610/aiounifi/pull/641

@mkmer I think this also does exactly the same, what do you say?


mkmer:
It does solve the issue I was observing (in a cleaner more readable way too)

PR #801, opened 2025-01-20
Add support for optional 'for_hotspot' parameters in login process

sir-Unknown:
To enable hotspot operator user to login.

This update introduces the ability to include the 'for_hotspot' and 'site_name' fields in the login request if they are provided in the configuration. 

Based on the following observed working example for logging in as a hotspot operator:
curl 'https://...:8443/api/login' --data-raw '{"username":"*******","password":"******","for_hotspot":true,"remember":false,"site_name":"default"}' --insecure

Adding the parameters "for_hotspot": true and "site_name": "default" resolves login issues for hotspot operator users.

See this for more information: https://github.com/ufozone/ha-unifi-voucher/issues/125#issuecomment-2601026022
Changes are verified to be working: https://github.com/ufozone/ha-unifi-voucher/issues/125#issuecomment-2602119647

### DIFF
--- a/aiounifi/__main__.py
+++ b/aiounifi/__main__.py
@@ -23,6 +23,7 @@ async def unifi_controller(
     site: str,
     session: aiohttp.ClientSession,
     ssl_context: SSLContext | None = None,
+    for_hotspot: bool | None = None, # New argument
 ) -> Controller | None:
     """Set up UniFi controller and verify credentials."""
     controller = Controller(
@@ -33,6 +34,7 @@ async def unifi_controller(
             password=password,
             port=port,
             site=site,
+            for_hotspot=for_hotspot, # New argument
             ssl_context=ssl_context if ssl_context is not None else False,
         )
     )
--- a/aiounifi/interfaces/connectivity.py
+++ b/aiounifi/interfaces/connectivity.py
@@ -72,10 +72,10 @@ class Connectivity:
         }
 
         if hasattr(self.config, "site") and self.config.site is not None:
-            auth["siteName"] = self.config.site
+            auth["site_name"] = self.config.site
 
         if hasattr(self.config, "for_hotspot") and self.config.for_hotspot is not None:
-            auth["forHotspot"] = self.config.for_hotspot
+            auth["for_hotspot"] = self.config.for_hotspot
         
         response, bytes_data = await self._request("post", url, json=auth)
 
--- a/aiounifi/interfaces/connectivity.py
+++ b/aiounifi/interfaces/connectivity.py
@@ -71,6 +71,12 @@ class Connectivity:
             "rememberMe": True,
         }
 
+        if hasattr(self.config, "site") and self.config.site is not None:
+            auth["siteName"] = self.config.site
+
+        if hasattr(self.config, "for_hotspot") and self.config.for_hotspot is not None:
+            auth["forHotspot"] = self.config.for_hotspot
+        
         response, bytes_data = await self._request("post", url, json=auth)
 
         if response.content_type != "application/json":
--- a/aiounifi/models/configuration.py
+++ b/aiounifi/models/configuration.py
@@ -18,6 +18,7 @@ class Configuration:
     password: str
     port: int = 8443
     site: str = "default"
+    for_hotspot: bool | None = None
     ssl_context: SSLContext | Literal[False] = False
 
     @property


aiounifi/interfaces/connectivity.py
```diff
@@ -71,6 +71,12 @@ async def login(self) -> None:
             "rememberMe": True,
         }
 
+        if hasattr(self.config, "site") and self.config.site is not None:
```
Kane610, line 74: site will always exist as part of the configuration class and it can only be a string so the if-statement is unnecessary.

aiounifi/interfaces/connectivity.py
```diff
@@ -71,6 +71,12 @@ async def login(self) -> None:
...
+            auth["site_name"] = self.config.site
```
Kane610, line 75: Is this actually needed for the hotspot? All requests contain the site as part of the path.

aiounifi/models/configuration.py
```diff
@@ -18,6 +18,7 @@ class Configuration:
     password: str
     port: int = 8443
     site: str = "default"
+    for_hotspot: bool | None = None
     ssl_context: SSLContext | Literal[False] = False
```
Kane610, line 22:
```suggestion
    ssl_context: SSLContext | Literal[False] = False
    for_hotspot: bool = False
    # Comment here about what this is
```
No need for None in this case
Please add a description to this as to what it does

aiounifi/interfaces/connectivity.py
```diff
@@ -71,6 +71,12 @@ async def login(self) -> None:
...
+
+        if hasattr(self.config, "for_hotspot") and self.config.for_hotspot is not None:
+            auth["for_hotspot"] = self.config.for_hotspot
```
Kane610, line 78:
```suggestion
        if self.config.for_hotspot:
            auth["for_hotspot"] = True
```
Based on other review comment we for_hotspot can only be a bool and then the only valid state to send the for_hostpot is when it is true if I understand its use case

Kane610 (changes requested):
Could you please describe this hotspot operator a bit, what is it allowed to do. Are there other considerations in the library that needs to be taken into account when for_hotspot is enabled? Maybe for_hotspot is good a name for a key when sending it to unifi but are there a better name for it in the library?

sir-Unknown:
> Could you please describe this hotspot operator a bit, what is it allowed to do. Are there other considerations in the library that needs to be taken into account when for_hotspot is enabled? Maybe for_hotspot is good a name for a key when sending it to unifi but are there a better name for it in the library?

As per the [Unifi help page](https://help.ui.com/hc/en-us/articles/1500011491541-UniFi-Roles-Explained-Admins-and-Users#:~:text=Hotspot%20Operator%20(Network)%3A%20Manage,multi%2Dsite%20UniFi%20Consoles) the hotspot operator has the following role: _Hotspot Operator (Network): Manage guest WiFi hotspots._ There isn't much more in the documentation about it in the official documentation. From what i've found and experienced a hotspot user is only able to generate and manage guest vouchers for the hotspot landing page. 

In the UniFi Network Application the hotspot operator is called "Hotspot Manager", so that might be a more friendly name.

aiounifi/interfaces/connectivity.py
sir-Unknown: Yes, when testing the authentication without this part, the login failed.

aiounifi/interfaces/connectivity.py
sir-Unknown: For the authentication as a hotspot manager to succeed, it is necessary to include both "site_name=[default]" and "for_hotspot=true".

aiounifi/interfaces/connectivity.py
```diff
@@ -71,6 +71,12 @@ async def login(self) -> None:
...
+
```
Kane610, line 79:
```suggestion
        if self.config.hotspot_manager_account:
            auth["for_hotspot"] = True
            auth["site_name"] = self.config.site
        
```
Allright, lets keep them together and only add them when its for a hotspot manager

aiounifi/__main__.py
```diff
@@ -23,6 +23,7 @@ async def unifi_controller(
     site: str,
     session: aiohttp.ClientSession,
     ssl_context: SSLContext | None = None,
+    for_hotspot: bool | None = None, # New argument
```
Kane610, line 26:
```suggestion
```
Either remove this or expand the argument parser inputs as well

aiounifi/__main__.py
```diff
@@ -33,6 +34,7 @@ async def unifi_controller(
             password=password,
             port=port,
             site=site,
+            for_hotspot=for_hotspot, # New argument
```
Kane610, line 37:
```suggestion
```
Either remove this or expand the argument parser inputs as well

Kane610 (changes requested):
> In the UniFi Network Application the hotspot operator is called "Hotspot Manager", so that might be a more friendly name.

Ok, lets go with `hotspot_manager_account` in that case, I think that makes it very clear

Please add a test case as well to validate that

PR #872, opened 2025-05-20
Add hide_ssid and use it for qr code generation

HerrHofrat:
I noticed that the QR codes generated in home assistant were not working for me as my SSIDs are hidden. I added the hide_ssid flag that is available in the unifi api and pass it down to the qr code generation.

Please let me know what you think - cheers


aiounifi/interfaces/wlans.py
```diff
@@ -28,4 +28,4 @@ async def disable(self, wlan: Wlan) -> TypedApiResponse:
 
     def generate_wlan_qr_code(self, wlan: Wlan) -> bytes:
         """Generate QR code based on WLAN properties."""
-        return wlan_qr_code(wlan.name, wlan.x_passphrase)
+        return wlan_qr_code(wlan.name, wlan.x_passphrase, hidden=wlan.hide_ssid if wlan.hide_ssid is not None else False)
```
Kane610, line 31:
```suggestion
        return wlan_qr_code(wlan.name, wlan.x_passphrase, hidden=wlan.hide_ssid is True)
```
can it be simplified to this

aiounifi/models/wlan.py
```diff
@@ -20,6 +20,7 @@ class TypedWlan(TypedDict):
     dtim_ng: int
     enabled: bool
     group_rekey: int
+    hide_ssid: bool
```
Kane610, line 23:
```suggestion
    hide_ssid: NotRequired[bool]
```
You're checking for None so I would expect this is needed

Kane610 (changes requested):
Thanks! Minor tweaks

aiounifi/interfaces/wlans.py
HerrHofrat: Thanks for the suggestion - I changed that in https://github.com/Kane610/aiounifi/pull/872/commits/6e62ec11320ba630d7724a07fe09b5b1ec836e9e

PR #936, opened 2025-09-26
Update DeviceSetPortEnabledRequest to match the GUI behavior.

deltaclock:
As of current, when `port_security_enabled` is set to `True` - the unifi controller will switch (and broadcast to the websocket) that both VLAN settings are disabled (`'native_networkconf_id': '', 'tagged_vlan_mgmt': 'block_all'`). This matches the behavior of the GUI.

On the invert of that (when the user needs to **enable** the port), since **only** the `port_security_enabled` property is modified - both VLANS settings will remain to the previous *disabled* mode. This deviates from the GUI, where there the VLAN id will be set to the configured default and tagged traffic will be set to auto.

A unique side effect of this also appears in the hass integration. When both these settings remain to the *disabled state*, the port on the `port_table` event will have it's `enabled` attribute to `false`. This creates a very weird state (not even allowed by the GUI) where the port is technically **enabled**, but since all VLAN traffic is **disabled**, hass thinks the whole port is disabled. https://github.com/home-assistant/core/blob/87f0703be17a20f0f87a1d7d76ca0781173555ab/homeassistant/components/unifi/switch.py#L371

Current code causes the port to be set to active (red box) while the VLANs stay disabled (green box). image

While this patch works (on my hass/switch), I think the ideal scenario would be to store the port configuration (somewhere/how/where?) *before it's disabled* and **revert** this when it's re-enabled. This idea, doesn't really concern this library directly, but could be a nice feature to have. (I could help there, but not really familiar on what's possible on hass)

### DIFF
--- a/aiounifi/models/device.py
+++ b/aiounifi/models/device.py
@@ -889,6 +889,7 @@ class Device(ApiItem):
 
     @property
     def connection_network_id(self) -> str:
+        """Native VLAN network ID."""
         return self.raw["connection_network_id"]
 
     @property
--- a/aiounifi/models/device.py
+++ b/aiounifi/models/device.py
@@ -892,7 +892,7 @@ class Device(ApiItem):
     @property
     def connection_network_id(self) -> str:
         """Native VLAN network ID."""
-        return self.raw["connection_network_id"]
+        return self.raw.get("connection_network_id", "")
 
     @property
     def considered_lost_at(self) -> int:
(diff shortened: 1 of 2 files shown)


Kane610:
Please update the tests.

I think the data can be stored similarly to how the UnifiWirelessClients class store data to track if a client is wireless or wired in the integration.

deltaclock:
Tests updated. It may require testing with actual devices, the test devices didn't contain any `connection_network_id` (as opposed to my device). Hopefully these extra fields are just ignored in this case.

jata1:
Hello @Kane610 - I just wanted to check in on this bug/PR. Any plans to finalise it and then release an update for the Unifi Network integration in HA? (I know you have been busy with a reno. Hope the project is going well :-))

PR #941, opened 2025-10-11
api_key support

ReSearchITEng:
API Key alternative to user/pass, on newer controllers that support it.


Kane610 (changes requested):
Will the api_key work with the existing api endpoints? I assumed it would only work with the new specific api

How I see it all api_key logic should be in handled connectivity apart from configuration definition, also add a new test to validate expected behavior

ReSearchITEng:
> Will the api_key work with the existing api endpoints? I assumed it would only work with the new specific api

I tested with same endpoints.
https://help.ui.com/hc/en-us/articles/30076656117655-Getting-Started-with-the-Official-UniFi-API :

There is the site level (HL overview): https://developer.ui.com/site-manager-api/gettingstarted

 And there is per **local** UniFI Application (probably the most used one for automations)

>  navigate to UniFi Network > Settings > Control Plane > Integrations. This includes a variety of capabilities including:
> UniFi devices (e.g., access points, switches, and gateways)
> Client activity (e.g., connected devices, usage statistics)
> Traffic insights (e.g., bandwidth consumption, network trends)

https://help.ui.com/hc/en-us/articles/30076656117655-Getting-Started-with-the-Official-UniFi-API

nberardi:
With the release of Fabric, it's getting really hard to add local user accounts.  Would anyone mind if I took an additional try at this approach, based on the learnings form this thread?

Kane610:
Of course. Though I don't know what fabric is

There are som unanswered questions like which apis this would be relevant for.

nberardi:
I opened a new PR that tries again from current master, using the review notes here: https://github.com/Kane610/aiounifi/pull/1088

The main differences from this branch:
- API keys only authenticate the official Integration API (`/proxy/network/integration/v1/...`), not classic `/api/login` or `/api/s/...`
- `X-API-KEY` lives in v1 connectivity, not on classic session headers
- Tests hit real `GET /v1/info` and `GET /v1/sites` behavior (header, skip login, 401)
- First endpoints that need the key are included (`/v1/info` and `/v1/sites`), per the later comment that a key without an API is not that useful

nberardi:
> Of course. Though I don't know what fabric is
> 
> There are som unanswered questions like which apis this would be relevant for.

It's a unified network solution that allows you to bind your different sites together, and once this happens you have a centrally managed identity server, which removes the ability to add local users.  

https://academy.ui.com/topics/zero-trust-networking-with-unifi-fabrics
https://help.ui.com/hc/en-us/articles/30979808349463-Getting-Started-with-UniFi-Fabrics

nberardi:
Holding off on that new PR for now — #1088 is closed until the approach is settled. Sorry for the noise.

Kane610:
For any value adding the api key some functionality should also be added.

I guess I should complete https://github.com/Kane610/aiounifi/pull/1010

PR #954, opened 2025-11-17
fix: Cast led_override_color_brightness from API to int

Sese-Schneider:
The device API sometimes returns the `led_override_color_brightness` as `str`.
I've now managed to observe this behavior in the wild on my setup with the debugger attached!

image
<br><br>

Adds a cast to `int` to ensure consistent datatypes for downstream consumers.

As discussed here: https://github.com/home-assistant/core/pull/155936#discussion_r2506351597

I can open a PR to HA once the fixed version is available.


Kane610:
Thanks! I'll try to do a release tomorrow

PR #1020, opened 2026-04-26
Add speedtest interface

esand:
This exposes UniFI speedtest results by calling the (undocumented?) `/speedtest` API endpoint. It also allows triggering a speedtest using the `/cmd/devmgr/speedtest` API endpoint. This exposes only the most recent speedtest result from the API call. Dual-WAN support at this time is *unknown*.

This only supports newer controllers that have these API endpoints. The (now "old") stand-alone UniFi Network Application doesn't support these newer API endpoints. Works with the new UniFi OS Server and I assume other cloud gateway hardware appliances that aren't EOL.

This will allow exposing speedtest results in Home Assistant's native UniFi integration.

### DIFF
--- a/aiounifi/models/speedtest.py
+++ b/aiounifi/models/speedtest.py
@@ -3,23 +3,21 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-from typing import TypedDict
+from typing import Any, NotRequired, TypedDict
 
 from .api import ApiItem, ApiRequest, ApiRequestV2, TypedApiResponse
 
 
-class TypedSpeedtestStatus(TypedDict, total=False):
+class TypedSpeedtestStatus(TypedDict):
     """Speedtest status type definition."""
 
-    status: str
-    status_text: str
     download_mbps: float
     upload_mbps: float
     latency_ms: float
     time: int
-    id: str
-    interface_name: str
-    wan_networkgroup: str
+    id: NotRequired[str]
+    interface_name: NotRequired[str]
+    wan_networkgroup: NotRequired[str]
 
 
 @dataclass
@@ -35,8 +33,18 @@ class SpeedtestStatusRequest(ApiRequestV2):
         """Decode response and extract nested data if present."""
         data = super().decode(raw)
 
-        if "data" in data and data["data"] and "data" in data["data"][0]:
-            data["data"] = data["data"][0]["data"]
+        if data.get("data"):
+            latest_by_interface: dict[str, dict[str, Any]] = {}
+            for result in data["data"]:
+                interface = result.get("interface_name", "default")
+                result["interface_name"] = interface
+
+                if interface not in latest_by_interface or result.get(
+                    "time", 0
+                ) >= latest_by_interface[interface].get("time", 0):
+                    latest_by_interface[interface] = result
+
+            data["data"] = list(latest_by_interface.values())
 
         return data
 
@@ -55,17 +63,6 @@ class SpeedtestStatus(ApiItem):
     """Represents a speedtest status."""
 
     raw: TypedSpeedtestStatus
-
-    @property
-    def status(self) -> str:
-        """Status of the speedtest."""
-        val = self.raw.get("status_text", self.raw.get("status"))
-        if val is not None:
-            return str(val)
-        # V2 endpoints don't seem to return a status explicitly for completed historical runs
-        if "download_mbps" in self.raw:
-            return "Completed"
-        return "unknown"
 
     @property
     def download(self) -> float:
--- a/aiounifi/models/speedtest.py
+++ b/aiounifi/models/speedtest.py
@@ -78,7 +78,15 @@ class SpeedtestStatus(ApiItem):
     @property
     def status(self) -> str:
         """Status of the speedtest."""
-        return str(self.raw.get("status_text", self.raw.get("status", self.raw.get("speedtest_status", "unknown"))))
+        val = self.raw.get(
+            "status_text", self.raw.get("status", self.raw.get("speedtest_status"))
+        )
+        if val is not None:
+            return str(val)
+        # V2 endpoints don't seem to return a status explicitly for completed historical runs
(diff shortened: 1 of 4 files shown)


aiounifi/models/speedtest.py
```diff
@@ -0,0 +1,88 @@
...
+
+    status: str
+    status_text: str
+    download_mbps: float
+    upload_mbps: float
+    latency_ms: float
+    time: int
+    id: str
+    interface_name: str
+    wan_networkgroup: str
```
Kane610: Do you know which ones will not always be reported? I prefer NotRequired rather than totalt=False when possible

Kane610 (changes requested):
Make sure to run all the pre-commit tests,

thecode:
> Dual-WAN support at this time is unknown.

The UI does separate speedtest per WAN, I think it should be supported

esand:
> > Dual-WAN support at this time is unknown.
> 
> The UI does separate speedtest per WAN, I think it should be supported

I don't have a second connection to test with. Are you able to give me the data output of `curl -k -X GET 'https://<UNIFI_IP>:11443/proxy/network/v2/api/site/default/speedtest' -H 'X-API-KEY: <API KEY HERE>' -H 'Accept: application/json' > output.json` (generate an API key in your integrations).

aiounifi/models/speedtest.py
esand: I'll go through my data and double check. I *think* status/status_text are likely not being returned and can be removed entirely (and the code that even reports a status). I'll circle back on this.

The issue I have is that this API endpoint seems totally undocumented, so I wouldn't know what non-standard situations to even handle unless I encounter them or someone reports them.

All that said however, I could always remove `total=False` and switch to NotRequired for the non-essentials such as `id`, `interface_name`, and `wan_networkgroup` since I don't use those.

Kane610:
HA has already been branched for 2026.5 so let's continue with this initial scope. easy to expand when we know more

esand:
I cleaned up the status properties since they were fake and added what I hope is going to allow for multi-wan support. I assumed that all the speedtest results come in as a single array of results, but the different ports would be identified in the `interface_name`. It should now go through and grab the most recent result per `interface_name` and key the results with it as well.

thecode:
> I don't have a second connection to test with. Are you able to give me the data output of `curl -k -X GET 'https://<UNIFI_IP>:11443/proxy/network/v2/api/site/default/speedtest' -H 'X-API-KEY: <API KEY HERE>' -H 'Accept: application/json' > output.json` (generate an API key in your integrations).

```json
{
    "data": [{
            "download_mbps": 12,
            "id": "69c05aa3c5b2302e963ab446",
            "interface_name": "eth8",
            "latency_ms": 34,
            "time": 1774213795000,
            "upload_mbps": 2,
            "wan_networkgroup": "WAN2"
        }, {
            "download_mbps": 933,
            "id": "69d418e5d47b2030ff6fcc22",
            "interface_name": "ppp0",
            "latency_ms": 4,
            "time": 1775507685000,
            "upload_mbps": 250,
            "wan_networkgroup": "WAN"
        }
    ]
}
```

esand:
@thecode - thank you. It seems that data format is what I had assumed it would be like... the latest updates should allow it to pick up both interfaces and report both of them.

aiounifi/models/speedtest.py
```diff
@@ -0,0 +1,88 @@
...
+                result["interface_name"] = interface
+
+                if interface not in latest_by_interface or result.get(
+                    "time", 0
+                ) >= latest_by_interface[interface].get("time", 0):
+                    latest_by_interface[interface] = result
+
+            data["data"] = list(latest_by_interface.values())
+
+        return data
```
Kane610, line 52: This is a lot more logic than usually in the decode function, is this warranted?

Kane610 (changes requested):
aiounifi/models/speedtest.py                       49      1    98%   38

One open question and a miss on test coverage. Nearly there, good job!

aiounifi/models/speedtest.py
esand: So, the .decode() function tosses the result into a data array... and the result we get from the API already has a data array, so we have a data array nested in a data array which is the first few lines.

The remaining lines are to support multi-wan. The results are in a single array and each object is a speedtest result for a particular interface. We have to go through each one and check the interface name and the timestamp when the result is from and ensure we get the latest result for each unique interface name.

The reason it's more logic than before (or perhaps usual?) is because of the multi-wan which I didn't have before, and because of the result format being a single list meaning we have to iterate all of it (unless we want to assume that the latest results are always at the top).

esand:
> aiounifi/models/speedtest.py 49 1 98% 38

Shoot - got so caught up removing needless stuff and doing multi-wan that I forgot to check for test coverage. I'll check what's missing and add it in.

Kane610:
No worries

esand:
Added missing test coverage for 100%.

aiounifi/models/speedtest.py
Kane610: Any reasoning on this

esand: I tried to submit a comment to this several days ago; seems to have got stuck in Pending status, so here it is again:

So, the super().decode() function tosses the result into a data array... and the result we get from the API already has a data array, so we have a data array nested in a data array which is the first few lines.

The remaining lines are to support multi-wan. The results are in a single array and each object is a speedtest result for a particular interface. We have to go through each one and check the interface name and the timestamp when the result is from and ensure we get the latest result for each unique interface name.

The reason it's more logic than before (or perhaps usual?) is because of the multi-wan which I didn't have before, and because of the result format being a single list meaning we have to iterate all of it (unless we want to assume that the latest results are always at the top). It used to be easier, just sorting it by time and picking the latest one... but now we have to consider two keys.

Kane610 (approved):

PR #1022, opened 2026-05-03
Add object-oriented network config support

ryanr14:
## Summary

Add support for UniFi Object-Oriented Network configuration objects, which are used by newer Policy Engine rules in the UniFi Network application.

This adds:
- a handler for `object_oriented_network_configs`
- list support for `/object-oriented-network-configs`
- update/toggle support for `/object-oriented-network-config/{id}`
- model properties for the rule name, enabled state, target type, targets, and secure/qos/route config blocks

## Context

Home Assistant can already expose legacy traffic rules/routes and zone-based firewall policies, but the UniFi UI now creates some simple Policy Engine rules as OON configs with generated firewall-policy rows underneath them. Those generated rows are marked predefined and are not the right object to update directly.

## Tests

- `python -m ruff check aiounifi tests pyproject.toml`
- `python -m pytest`

### DIFF
--- a/aiounifi/interfaces/api_handlers.py
+++ b/aiounifi/interfaces/api_handlers.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 from abc import ABC
 from collections.abc import Callable, ItemsView, Iterator, ValuesView
 import enum
-from typing import TYPE_CHECKING, Any, Generic, final
+from typing import TYPE_CHECKING, Any, Generic, cast, final
 
 from ..models.api import ApiItemT, ApiRequest
 
@@ -110,6 +110,17 @@ class APIHandler(SubscriptionHandler, Generic[ApiItemT]):
         for raw_item in raw:
             self.process_item(raw_item)
 
+    def _obj_id_from_raw(self, raw: dict[str, Any]) -> str | None:
+        """Return object ID from raw data."""
+        if self.obj_id_key in raw:
+            return cast(str, raw[self.obj_id_key])
+
+        if self.alt_obj_id_key is None or self.alt_obj_id_key not in raw:
+            return None
+
+        raw[self.obj_id_key] = raw[self.alt_obj_id_key]
+        return cast(str, raw[self.obj_id_key])
+
     @final
     def process_message(self, message: Message) -> None:
         """Process and forward websocket data."""
@@ -122,17 +133,10 @@ class APIHandler(SubscriptionHandler, Generic[ApiItemT]):
     @final
     def process_item(self, raw: dict[str, Any]) -> None:
         """Process item data."""
-        obj_id_key = self.obj_id_key
-        if obj_id_key not in raw:
-            if self.alt_obj_id_key is None or self.alt_obj_id_key not in raw:
-                return
-            obj_id_key = self.alt_obj_id_key
+        if (obj_id := self._obj_id_from_raw(raw)) is None:
+            return
 
-        if self.obj_id_key not in raw:
-            raw[self.obj_id_key] = raw[obj_id_key]
-
-        obj_id: str
-        obj_is_known = (obj_id := raw[obj_id_key]) in self._items
+        obj_is_known = obj_id in self._items
         self._items[obj_id] = self.item_cls(raw)
 
         self.signal_subscribers(
@@ -143,8 +147,10 @@ class APIHandler(SubscriptionHandler, Generic[ApiItemT]):
     @final
     def remove_item(self, raw: dict[str, Any]) -> None:
         """Remove item."""
-        obj_id: str
-        if (obj_id := raw[self.obj_id_key]) in self._items:
+        if (obj_id := self._obj_id_from_raw(raw)) is None:
+            return
+
+        if obj_id in self._items:
             self._items.pop(obj_id)
             self.signal_subscribers(ItemEvent.DELETED, obj_id)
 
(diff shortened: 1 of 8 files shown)


aiounifi/interfaces/api_handlers.py
```diff
@@ -83,6 +83,7 @@ class APIHandler(SubscriptionHandler, Generic[ApiItemT]):
     """Base class for a map of API Items."""
 
     obj_id_key: str
+    alt_obj_id_key: str | None = None
```
Kane610:
```suggestion
obj_id_key: str | tuple[str]
"""Key for object ID, if tuple left to right in priority order is expected."""
```
Lets not introduce multiple variations

aiounifi/interfaces/api_handlers.py
```diff
@@ -121,11 +122,17 @@ def process_message(self, message: Message) -> None:
...
         """Process item data."""
+        obj_id_key = self.obj_id_key
+        if obj_id_key not in raw:
+            if self.alt_obj_id_key is None or self.alt_obj_id_key not in raw:
+                return
+            obj_id_key = self.alt_obj_id_key
+
         if self.obj_id_key not in raw:
-            return
+            raw[self.obj_id_key] = raw[obj_id_key]
```
Kane610:
```suggestion
       obj_id_key: str | None = None
        if type(self.obj_id_key) is str:
            if obj_id_key in raw:
                obj_id_key = self.obj_id_key
        else:
            obj_id_key = next((k for k in self.obj_id_key if k in raw), None)
        if obj_id_key is None:
                return
```
Then it would be something like this

aiounifi/models/object_oriented_network_config.py
```diff
@@ -0,0 +1,121 @@
...
+
+    @property
+    def enabled(self) -> bool:
+        """Is object-oriented network configuration enabled."""
+        return self.raw["enabled"]
+
+    @property
+    def target_type(self) -> str:
+        """Target type for object-oriented network configuration."""
+        return self.raw.get("target_type", "CLIENTS")
```
Kane610: Why should default be CLIENTS?

aiounifi/models/object_oriented_network_config.py
```diff
@@ -0,0 +1,121 @@
...
+
+class TypedObjectOrientedNetworkConfig(TypedDict):
+    """Object-oriented network configuration type definition."""
+
+    _id: NotRequired[str]
+    id: NotRequired[str]
+    enabled: bool
+    name: str
+    target_type: NotRequired[str]
+    targets: NotRequired[list[Any]]
```
Kane610: Looking at the mock data this can be a list[str | dict[str, str]] or something

pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name        = "aiounifi"
-version     = "90"
+version     = "91"
```
Kane610: Should not be part of a feature PR

Kane610 (changes requested):

aiounifi/models/object_oriented_network_config.py
```diff
@@ -0,0 +1,124 @@
...
+from typing import Any, NotRequired, Self, TypedDict
+
+from .api import ApiItem, ApiRequestV2
+
+
+class ObjectOrientedNetworkInternet(TypedDict):
+    """Internet access configuration."""
+
+    mode: str
+    schedule: NotRequired[dict[str, Any]]
```
Kane610:
```suggestion
    schedule: NotRequired[dict[str, str]]
```

aiounifi/interfaces/api_handlers.py
ryanr14: Good call. I changed this so `obj_id_key` is the one place a handler describes its ID fields. It can now be a string or an ordered tuple, so OON uses `("_id", "id")` without adding a second config knob. That keeps the base handler smaller and makes the priority explicit.

aiounifi/interfaces/api_handlers.py
ryanr14: I adjusted this in the same pass. `_obj_id_from_raw()` now treats `obj_id_key` as one-or-many keys and picks the first one present, and both `process_item()` and `remove_item()` use it. I also stopped having the generic handler rewrite `id` into `_id`; that felt a little too clever for a shared helper. The model/request layer already knows how to work with either shape.

aiounifi/models/object_oriented_network_config.py
ryanr14: Agreed. I removed the `CLIENTS` fallback. `target_type` now returns `str | None`, which is more honest about the API payload. If the controller does not send a value, aiounifi should not invent one.

aiounifi/models/object_oriented_network_config.py
ryanr14: Good catch. I tightened this to `list[str | ObjectOrientedNetworkTarget]`, with `ObjectOrientedNetworkTarget` covering the `{type, value}` shape. That matches the two formats we have seen without leaving the whole property as `Any`.

pyproject.toml
ryanr14: Yep, agreed. I reverted the version back to `90`. This PR should just carry the feature work and let the normal release flow choose the next package version.

aiounifi/models/object_oriented_network_config.py
ryanr14: Yep, tightened this to `NotRequired[dict[str, str]]`. The schedule data we have here is string-key/string-value, so `Any` was wider than it needed to be.

Kane610:
I'll get back to this within the next day or two!

Kane610 (approved):
Sweet!

ryanr14:
Thanks for the review and merge!

PR #1027, opened 2026-05-10
fix: handle missing name field in FirewallPolicy

AZSD1-6:
## Problem
Some predefined/system UniFi firewall policies omit the name field from the API response, causing a KeyError: name crash when Home Assistant loads the UniFi integration.

## Fix
- Mark name as NotRequired in TypedFirewallPolicy
- Use .get('name') returning str | None, matching the existing pattern for description

## Reproduction
Any UniFi network with predefined zone-based firewall policies triggers this in the HA config_entries log.

### DIFF
--- a/aiounifi/models/firewall_policy.py
+++ b/aiounifi/models/firewall_policy.py
@@ -29,7 +29,7 @@ class TypedFirewallPolicy(TypedDict):
 
     _id: str
     action: str
-    name: str
+    name: NotRequired[str]
     enabled: bool
     connection_state_type: str
     connection_states: list[str]
@@ -62,7 +62,7 @@ class FirewallPolicy(ApiItem):
     @property
     def name(self) -> str:
         """Firewall policy name."""
-        return self.raw["name"]
+        return self.raw.get("name", "")
 
     @property
     def enabled(self) -> bool:


aiounifi/models/firewall_policy.py
```diff
@@ -60,9 +60,9 @@ def id(self) -> str:
         return self.raw["_id"]
 
     @property
-    def name(self) -> str:
+    def name(self) -> str | None:
         """Firewall policy name."""
-        return self.raw["name"]
+        return self.raw.get("name")
```
Kane610:
```suggestion
    def name(self) -> str:
"""Firewall policy name."""
        return self.raw.get("name", "")
```
In most cases of name we typically keep it as a string

Kane610:
Once this is fixed those firewall policies will all have the same blank name, how would this be better managed in home assistant? are there other fields that could be useful?

aiounifi/models/firewall_policy.py
AZSD1-6: Accepted -- updated to return `str` with `""` default, consistent with the rest of the library.

AZSD1-6:
Good question. For unnamed predefined policies the blank name would be unhelpful in HA.

The most natural fallback would be to use description if available (already NotRequired[str] in the model), since predefined policies often have meaningful descriptions. If that is also blank, falling back to something like Policy {self.index} would at least give a stable unique label based on priority order.

Would you prefer I add a display_name helper property here in aiounifi that implements that fallback, or is it better handled on the HA integration side?

Kane610:
Im not sure where it should be, if there is information available from unifi it should be exposed and then it could be selected as fallback on HA side. Do you have information in mind to resolve this?

proscar87:
@Kane610 Not my PR, but your question about what could serve as a fallback looked answerable from the payload, so I went and checked.

First, the crash is still live on master — `firewall_policy.py:65` is `self.raw["name"]` with `name: str` required in the TypedDict:

```
CRASH reproducido: KeyError('name')
```

On what's left to identify a policy with no name: quite a lot, and none of it needs a new field from this library. Dropping `name` from the fixture policy leaves

```
action                = 'ALLOW'
predefined            = True
index                 = 10000
ip_version            = 'BOTH'
protocol              = 'all'
connection_state_type = 'ALL'
source.zone_id        = '678c63bc…'
destination.zone_id   = '678ccc26…'
```

The natural identity of a zone-based policy is **action + source zone + destination zone** — that's how the UniFi UI itself presents predefined rules. Both zone IDs are already exposed on `FirewallPolicyEndpoint`, and `FirewallZone` already carries `name` and `zone_key`, so a consumer can resolve `ALLOW LAN → WAN` from data this library already returns today. `index` gives a stable tiebreaker when two predefined rules share that triple.

So my read is that nothing new needs exposing — the fallback can be composed downstream, which also keeps the naming policy in Home Assistant rather than baking a display string into the library.

One suggestion on the fix itself, and it's the reason I'm commenting rather than just 👍: `self.raw.get("name", "")` makes an unnamed policy indistinguishable from one a user named with an empty string, so the consumer can't tell when to apply the fallback. `str | None` via `self.raw.get("name")` preserves that signal and matches the existing `description` pattern right below it. @AZSD1-6's diff already marks the key `NotRequired`, so it'd be a one-character change to the return.

Unrelated, but it'd bite anyone writing a test for the above: the fixtures don't cross-reference. `FIREWALL_POLICIES[0]["source"]["zone_id"]` is `678c63bc2d97692f08adcdfa`, while the only similar zone in `FIREWALL_ZONES` is `678c63bc2d97692f08adcdfb` (WAN) — one character apart, and no zone carries the `…cdfa` id. Harmless today since nothing joins them, but zone resolution can't be tested against those fixtures as they stand.

Happy to leave this with @AZSD1-6 — just wanted the question answered with the payload rather than from memory.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Kane610:
Thanks @proscar87  would like to complete it here and also in the integration?

proscar87:
Happy to help complete it — here's a concrete plan grounded in the actual model (checked `aiounifi/models/firewall_policy.py` on master):

**In aiounifi (this PR):** `TypedFirewallPolicy` already carries `description: NotRequired[str]` and a required `index: int`. A `name` property fallback chain of `raw["name"] or raw.get("description") or f"Policy {raw['index']}"` gives every policy a stable, non-blank label without inventing data — `index` is UniFi's own ordering field, always present, so it survives controller restarts and firmware updates rather than being derived client-side.

**In home-assistant/core's unifi integration:** as long as the firewall-policy entities there read `.name` off this model rather than `raw["name"]` directly, no separate code change should be needed there once this bump lands — just the routine `aiounifi` version bump PR. I'll check that assumption against the actual entity code before claiming it's a non-issue.

If that shape looks right to you, I can push the fallback + a test (predefined policy fixture with no `name`, asserting the description/index fallback order) to this branch, or open a follow-up PR here if you'd rather keep this one as-is — whichever you prefer. Let me know and I'll get it done.

Kane610:
Is it only specific rules which does not have a name? Is it internal rules or can any rule be created with a name, I would like to avoid putting too much other information into the name field, I would rather have the integration manage that. Maybe rules without names should not be created in home assistant?

proscar87:
Worth being precise about provenance here: the \"predefined rules lack a name\" claim is @AZSD1-6's original field report from their own controller — the actual crash they hit. What I verified was the code-level mechanism (dropping \`name\` from the one real fixture in this repo does reproduce \`KeyError('name')\`), not an independent survey of which real UniFi rules omit it. That fixture itself (\`FIREWALL_POLICIES[0]\` in \`tests/fixtures.py\`) is \`predefined: False\` with a real name — it doesn't reproduce the bug as-is, I only used it as a base to test the missing-key case. So the honest answer to "is it only specific/internal rules" is: I don't know from data, only from @AZSD1-6's report that *some* predefined policy on their controller had no name. They're the one who can say whether it's every predefined rule, some category of them, or something else — worth pinging them directly on that.

On keeping naming out of the library — agreed, and it matches what I said in my last comment too ("keeps the naming policy in Home Assistant rather than baking a display string into the library"). Concretely that'd mean: \`name\` stays a thin passthrough (\`self.raw.get("name")\`, typed \`str | None\` rather than the current required \`str\`) so a missing name is distinguishable from an empty string, and the integration decides what to do with \`None\` — compose a fallback, or per your last idea, just not create an entity for it. That second option sidesteps the naming question entirely, at the cost of predefined rules becoming invisible/uncontrollable in HA even though the firewall is still enforcing them — which may be exactly the right call if those rules aren't meant to be user-facing anyway, but that's a product call for you two, not something the library needs an opinion on.

Kane610 (approved):
Thanks

PR #1049, opened 2026-06-27
Expose typed OON secure config

ryanr14:
This is the smaller model change requested from home-assistant/core#169675.

Home Assistant needs to decide whether an object-oriented network config is a secure internet-blocking config without reaching into raw payload data. This exposes that through typed model properties:

- `config.secure.available` is false when `secure` is missing or null, such as route-only configs
- `config.secure.enabled` exposes the secure rule enabled flag with a safe default
- `config.secure.internet` is `None` when no internet config is present
- `config.secure.internet.mode` exposes `TURN_OFF_INTERNET` as `ObjectOrientedNetworkInternetMode.TURN_OFF_INTERNET`

Unknown internet modes map to `ObjectOrientedNetworkInternetMode.UNKNOWN` so new UniFi modes do not break parsing.

Tests:
- `.venv/bin/python -m pytest -o addopts=''`
- `.venv/bin/python -m pytest -o addopts='' --cov-report term-missing --cov=aiounifi.models.object_oriented_network_config tests/test_object_oriented_network_configs.py`
- `.venv/bin/python -m ruff check aiounifi/models/object_oriented_network_config.py tests/test_object_oriented_network_configs.py`
- `.venv/bin/python -m ruff format --check aiounifi/models/object_oriented_network_config.py tests/test_object_oriented_network_configs.py`
- `.venv/bin/python -m mypy aiounifi/models/object_oriented_network_config.py`

### DIFF
--- a/aiounifi/models/object_oriented_network_config.py
+++ b/aiounifi/models/object_oriented_network_config.py
@@ -104,7 +104,7 @@ class TypedObjectOrientedNetworkConfig(TypedDict):
     targets: NotRequired[list[str | ObjectOrientedNetworkTarget]]
     qos: NotRequired[ObjectOrientedNetworkQos]
     route: NotRequired[ObjectOrientedNetworkRoute]
-    secure: NotRequired[ObjectOrientedNetworkSecureT | None]
+    secure: NotRequired[ObjectOrientedNetworkSecureT]
 
 
 @dataclass
--- a/aiounifi/models/object_oriented_network_config.py
+++ b/aiounifi/models/object_oriented_network_config.py
@@ -22,7 +22,7 @@ class ObjectOrientedNetworkInternetMode(StrEnum):
 class ObjectOrientedNetworkSecureInternetT(TypedDict):
     """Internet access configuration."""
 
-    mode: ObjectOrientedNetworkInternetMode
+    mode: NotRequired[str]
     schedule: NotRequired[dict[str, str]]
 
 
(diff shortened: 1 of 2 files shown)


aiounifi/models/object_oriented_network_config.py
```diff
@@ -51,7 +104,7 @@ class TypedObjectOrientedNetworkConfig(TypedDict):
     targets: NotRequired[list[str | ObjectOrientedNetworkTarget]]
     qos: NotRequired[ObjectOrientedNetworkQos]
     route: NotRequired[ObjectOrientedNetworkRoute]
-    secure: NotRequired[ObjectOrientedNetworkSecure]
+    secure: NotRequired[ObjectOrientedNetworkSecureT | None]
```
Kane610:
```suggestion
    secure: NotRequired[ObjectOrientedNetworkSecureT]
```
When using NotRequired None should not be needed as an option?

Kane610 (changes requested):
Excellent, one observation

aiounifi/models/object_oriented_network_config.py
ryanr14: Good catch, thanks. That `| None` was unnecessary, so I removed it from the raw `TypedDict` and updated the related test to cover the missing-key case instead.

Kane610 (approved):

Kane610:
Awesome!

PR #1063, opened 2026-08-10
Recognise pasv24 as a real PoE mode

### LINKED ISSUE #1034: Unsupported port PoE mode passv24 is spamming HA logs
It seems that this class 

https://github.com/Kane610/aiounifi/blob/4e1ab2dc007880a6616860dc3038272d6830e87e/aiounifi/models/port.py#L31

does not support pasv24, which is fine.  However it's constantly spamming the logs every 30 seconds.

proscar87:
Fixes #1034.

`PortPoEMode` lists `auto`, `off`, `passthrough` and `unknown`. Passive 24V is missing, so `pasv24` falls through `_missing_` on every poll and logs a warning per affected port. One reporter in that issue sees roughly 200 per hour across five switches.

It isn't a setting users can move away from — UniFi Protect G3 cameras accept nothing but 24V passive — so "switch the port to PoE+" resolves it for one reporter and not the others.

The value is well evidenced in the issue: a full `port_table` entry with `"poe_mode": "pasv24"`, `"poe_voltage": "24.00"`, `"poe_caps": 7`, plus the matching `port_overrides` entry. And the library already documents the capability — `Port.poe_caps` carries `7 - 24V passive` in its docstring — so only the mode enum was behind.

## Change

One enum member, `PASSIVE_24V = "pasv24"`, plus the `poe_mode` docstring.

## Tests

`pasv24` joins the `test_port_poe_mode_enum_mapping` parameters, and a new test asserts it produces no log output at all — the warning is the actual complaint, so it's worth asserting on directly rather than inferring from the enum value.

One thing worth flagging: `test_port_poe_mode_unknown_fallback` used `"pasv24"` as its example of an unrecognised value. Left alone it would have kept passing while asserting the opposite of this fix, so it moves to `"pasv54"`.

Full suite: 242 passed, coverage 97.11%. `ruff check`, `ruff format --check` and `mypy aiounifi` all clean. Removing the enum member alone makes the new tests fail, so they are load-bearing.

I don't have passive-24V hardware to confirm the warning disappears in situ — this is verified against the payloads in the issue and by test.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Kane610 (approved):
Thanks

Issue #1062, opened 2026-08-08
Add UPS 2U Pro (USPDA2B / type usp) device and outlet support


jimstrang:
## Summary

The new UniFi UPS 2U Pro is reported by UniFi Network as:

- Device type: `usp` (new device type from unifi here)
- Model: `USPDA2B`
- Eight individually controllable and metered outlets

Please add explicit support and regression coverage for this device type.

## Device payload

The device reports an `outlet_table` with eight entries. Each outlet includes:

- `index`
- `name`
- `relay_state`
- `cycle_enabled`
- `outlet_caps: 3`
- `outlet_voltage`
- `outlet_current`
- `outlet_power`
- `outlet_power_factor`

`outlet_caps: 3` denotes relay control plus metering.

## Requested changes

1. Add `usp` to `DeviceType`.
2. Add a sanitized `USPDA2B` / UPS 2U Pro fixture.
3. Add device parsing coverage for a `type: usp` payload.
4. Add outlet-handler coverage that verifies all eight outlets preserve their control and telemetry fields.
5. Keep existing `USPPDUP` behavior unchanged.

I own a UPS 2U Pro and can take a pass at the implementation and provide a sanitized controller payload if useful.

I recently added equivalent support to Unpoller's UniFi library. The change classifies `USPDA2B` through the existing PDU path and adds a fixture plus tests for all eight outlet measurements. This should be a very similar change:

- https://github.com/unpoller/unpoller/issues/1047
- https://github.com/unpoller/unpoller/pull/1048
- https://github.com/unpoller/unifi/pull/233

## Acceptance criteria

A `USPDA2B` payload with `type: usp` produces a normal `Device` and eight `Outlet` objects, with relay state, capabilities, voltage, current, power, and power factor retained.

jimstrang:
@Kane610 I have sanitized controller payloads for the new UPS 2U Pro and related SmartPower devices, and would appreciate guidance on two API-shape questions before I open a PR.

The new UPS 2U Pro (`USPDA2B`, `type: "usp"`) differs from the existing PDU fixtures in two relevant ways:

| Device / outlet type | `outlet_caps` | Measurements |
| --- | ---: | --- |
| USP PDU Pro USB outlets | `1` | absent |
| USP PDU Pro AC outlets | `3` | JSON strings |
| UPS 2U (`USWDA25`, `type: "usw"`) | `65549` / `65541` | absent |
[...]

PR #1068, opened 2026-08-13
feat: support USP devices and outlet capabilities

jimstrang:
Adds support for UniFi devices reported as `type: "usp"`.

- Adds the UPS 2U Pro (`USPDA2B`) fixture and regression coverage
- Supports numeric outlet telemetry alongside existing string values
- Exposes relay and metering capabilities from outlet bitmasks

Related to #1062 and home-assistant/core#178529.


aiounifi/models/outlet.py
```diff
@@ -25,7 +34,11 @@ def has_relay(self) -> bool | None:
 
         Not reported by USP-PDU-Pro, see caps.
         """
-        return self.raw.get("has_relay")
+        if (has_relay := self.raw.get("has_relay")) is not None:
+            return has_relay
+        if self.caps is None:
+            return None
+        return bool(self.caps & OutletCapability.RELAY)
```
Kane610: just a small preference is to return None at the very last as that is of least relevance to the function

aiounifi/models/outlet.py
```diff
@@ -46,7 +59,11 @@ def has_metering(self) -> bool | None:
         Reported false by UP1 and UP6 which does not have power metering.
         Not reported by by USP-PDU-Pro, see caps.
         """
-        return self.raw.get("has_metering")
+        if (has_metering := self.raw.get("has_metering")) is not None:
+            return has_metering
+        if self.caps is None:
+            return None
+        return bool(self.caps & OutletCapability.METERING)
```
Kane610: just a small preference is to return None at the very last as that is of least relevance to the function

aiounifi/models/outlet.py
jimstrang: makes sense. will fix both.

Kane610:
Thanks!

PR #1087, opened 2026-09-10
feat: add WAN3 to device uptime stats type

sdelmas:
## What

`TypedDeviceUptimeStats` declares `WAN` and `WAN2` only, so a gateway with a
third uplink has its `uptime_stats.WAN3` block typed away. This adds it as
`NotRequired`, matching the way `TypedDeviceUptimeStatsWanMonitor.latency_average`
is already optional.

```python
 class TypedDeviceUptimeStats(TypedDict):
     """Device uptime stats type definition."""

     WAN: TypedDeviceUptimeStatsWan
     WAN2: TypedDeviceUptimeStatsWan
+    WAN3: NotRequired[TypedDeviceUptimeStatsWan]
```

Typing only — no runtime behaviour changes, since `uptime_stats` is a plain dict
at runtime and `.get("WAN3")` already worked. What it unblocks is the consumer
side: Home Assistant's UniFi integration hardcodes `("WAN", "WAN2")` and cannot
open that up while the type says a third WAN does not exist.

## Why NotRequired

A gateway reports only the interfaces it actually has, so `WAN3` is absent on
most hardware. `NotRequired` says exactly that, and keeps mypy honest at the call
site rather than promising a key that will not be there.

`WAN2` is arguably `NotRequired` too — a single-WAN gateway reports no `WAN2` —
but changing an existing key's requiredness is a separate call and would ripple
into current consumers, so I have left it alone.

## Evidence

Taken from a live UDM (`UDMA6A8`, firmware `5.1.31.34074`) with three uplinks.
Its `uptime_stats` carries `WAN`, `WAN2` and `WAN3`, each with the same
three-monitor set:

```
WAN3  www.microsoft.com  availability 100.0  latency_average 41
WAN3  google.com         availability 100.0  latency_average 32
WAN3  1.1.1.1            availability 100.0  latency_average 16
```

The `GATEWAY_USG3` fixture gains a matching `WAN3` block, `test_device`'s
reference data is extended to match, and `test_update_stats` asserts the third
WAN parses with its three monitors.

## Checks

- `pytest`: 246 passed, coverage 97.49% (gate is 95%)
- `ruff check .`: all checks passed
- `mypy aiounifi`: no issues found in 49 source files

## Note

A release including this would let me open the matching Home Assistant core PR,
which adds the WAN3 latency sensors. Happy to hold or reshape this if you would
rather approach the WAN keys differently.

---

AI assistance disclosure: this change was written with
[Claude Code](https://claude.com/claude-code), as recorded by the `Co-Authored-By`
trailer on the commit.


Kane610:
Thanks! I will make a new release during the weekend

PR #1088, opened 2026-09-11
Add UniFi Network Integration API key authentication

nickberardi:
## Summary

This is a new, smaller alternative to #941 (not a continuation of that branch).

UniFi Network API keys authenticate the **official Integration API**, not classic `/api/login` or `/api/s/{site}/...`. This PR adds key-only login plus the first Integration endpoints that actually need the key: `GET /v1/info` and `GET /v1/sites`.

Classic cookie/CSRF APIs, TOTP, and websocket behavior are unchanged. Username/password remain required for devices, clients, and events.

## Learnings from #941

Kane610's review asked three things this PR is built around:

1. **Keys are for the new API only.** `X-API-KEY` is not attached to classic `Connectivity.headers` or `/api/login`. Integration calls go to `{url}/proxy/network/integration/v1/...` from `aiounifi/network/v1/connectivity.py`.
2. **Key HTTP lives in connectivity.** `Configuration.api_key` is just the field (default `""`). Username/password also default to `""` so key-only setups do not need dummy credentials. Mixed username/password + key still does classic cookie login; `Controller.network` can use the key independently.
3. **Tests assert real request behavior.** Login with a key skips `/api/login` and `GET`s `/v1/info` with `X-API-KEY`. Invalid keys raise `Unauthorized` from that same call. Sites/info tests check the header, query params, JSON body, 401 envelopes, and decode errors.

This also follows the later note on #941 that there is little point adding a key without an API that needs it. `/v1/info` validates the key; `/v1/sites` is the first list endpoint (UUID `id` + `internalReference`).

## Scope

- `Configuration.api_key`
- Key-only `login()` validates `GET /v1/info` and skips classic login
- `Controller.network` Integration API client (`get_info()`, `sites.list()` / `list_page()`, `resolve_site_uuid()`)
- Classic `/api` and websocket stay cookie/CSRF

Not in this PR: remaining Integration resources (clients, devices, etc.), Site Manager cloud API, or using a key against classic endpoints.

After this is released on PyPI, Home Assistant can add Protect-style user vs API-key config-flow modes without putting protocol/auth in the integration.


nberardi:
Closing this for now — I opened it before we had agreed to publish. The branch remains on the fork if we want to reopen after the design discussion.

PR #1092, opened 2026-09-20
Add network configuration (networkconf) support

tomwilkie:
Add read and write support for /rest/networkconf, covering the WAN entries that hold multi-WAN failover and load balancing settings.

Nearly all fields are NotRequired since WAN entries on a real controller carry very different key sets depending on WAN type. Updates require the full network object; the controller rejects partial payloads.


tomwilkie:
Background: I want to have home assistant control the failover / load balancing setting of my unifi network, so I need this plumbing before I can do the changes in HA.

aiounifi/models/network.py
```diff
@@ -0,0 +1,111 @@
...
+        """Create network list request."""
+        return cls(method="get", path="/rest/networkconf")
+
+
+@dataclass
+class NetworkUpdateRequest(ApiRequest):
+    """Request object for network update."""
+
+    @classmethod
+    def create(cls, network: TypedNetwork) -> Self:
```
Kane610: The only thing I'd like changed is to use the Network class as input here

Kane610 (commented):
Thanks for the PR, I have one comment, Id like to see changed

aiounifi/models/network.py
tomwilkie: Done! That ended up being a slightly bigger change.  If you want to hold off merging, I've almost finished the updates to the HA integration where I consume this API.

I'll let you know when that done (towards he end of week) and if any more changes fall out of that.

Kane610: Sure I can wait, I'll be occupied a couple of days but its probably not any issues getting all into HA before beta cutoff next wednesday

tomwilkie: Here a preview of the core changes: https://github.com/home-assistant/core/compare/dev...tomwilkie:homeassistant-core:unifi-wan-failover-config?expand=1

Its pretty big, so I need to see if I can split them into a series of smaller PRs.  But it works!

I think we can merge this PR now.

PR #1093, opened 2026-09-20
Fix TypedDeviceUptimeStatsWan under-declaring gateway uptime stats

tomwilkie:
A gateway reports alerting_monitors, availability, latency_average, time_period and uptime alongside monitors, and does not always send monitors at all: a GRE tunnelled 5G WAN omits it entirely. Add the missing fields and make monitors NotRequired.

A monitor that never got a response omits latency_average entirely, so that field stays NotRequired[int].

WAN2 was declared required in TypedDeviceUptimeStats but is absent whenever that WAN is unplugged; the same applies to WAN, so make both NotRequired.


Kane610 (approved):
Awesome! Thanks

